PR #16, opened 2019-09-27
Improve the Header Authentication Method

obelisk:
This adds more of the code needed to properly authenticate a user based on headers present added by a reverse proxy.

I'm wary of the CSRF token line that's commented out but I must just be missing something there. Also updating users always fails and I assume that's because the login flow never runs with header authentication.

The way this is designed to work is to be put behind a proxy that handles the authentication for you that then passes the request along with newly added trusted headers from the proxy.

Admin vs User level permissions are handled by checking for the presence of specified groups passed along from the proxy.

### DIFF
--- a/cmd/admin/auth.go
+++ b/cmd/admin/auth.go
@@ -106,6 +106,7 @@ func handlerAuthCheck(h http.Handler) http.Handler {
 			}
 		case settings.AuthHeaders:
 			username := r.Header.Get(headersConfig.TrustedPrefix + headersConfig.UserName)
+			email := r.Header.Get(headersConfig.TrustedPrefix + headersConfig.Email)
 			groups := strings.Split(r.Header.Get(headersConfig.TrustedPrefix+headersConfig.Groups), ",")
 		    fullname := r.Header.Get(headersConfig.TrustedPrefix + headersConfig.DisplayName)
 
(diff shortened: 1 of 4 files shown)


cmd/admin/auth.go
```diff
@@ -102,8 +103,45 @@ func handlerAuthCheck(h http.Handler) http.Handler {
...
 			}
 		case settings.AuthHeaders:
-			// Access always granted
-			h.ServeHTTP(w, r)
+			username := r.Header.Get(headersConfig.TrustedPrefix + headersConfig.UserName)
+			groups := strings.Split(r.Header.Get(headersConfig.TrustedPrefix+headersConfig.Groups), ",")
+
+			if username == "" {
+				log.Printf("A username is required to use this system.")
+				w.WriteHeader(http.StatusBadRequest)
```
javuto: In other authentication methods I have used a redirect instead of writing the header:
```
http.Redirect(w, r, forbiddenPath, http.StatusFound)
```
Do you think we should use `WriteHeader` instead?

cmd/admin/auth.go
```diff
@@ -102,8 +103,45 @@ func handlerAuthCheck(h http.Handler) http.Handler {
...
+					break
+				} else if group == headersConfig.UserGroup {
+					s["level"] = userLevel
+					// We can't break because we might still find a higher permission level
+				}
+			}
+
+			// This user didn't present a group that has permission to use the service
+			if _, ok := s["level"]; !ok {
+				w.WriteHeader(http.StatusForbidden)
```
javuto: Ditto

cmd/admin/auth.go
```diff
@@ -102,8 +103,45 @@ func handlerAuthCheck(h http.Handler) http.Handler {
...
+				return
+			}
+
+			err := adminUsers.UpdateMetadata(r.RemoteAddr, r.Header.Get("User-Agent"), username)
```
javuto: Since this authentication method does not require to pre-create users, better follow these steps:
1 - Check if user already exists
2 - If not, create user
3 - Update metadata
I would replace this block (135-138) with the following code:
```
if !adminUsers.Exists(username) {
  log.Printf("user not found, creating new user: %s", username)
  email := r.Header.Get(headersConfig.TrustedPrefix + headersConfig.Email)
  fullname := r.Header.Get(headersConfig.TrustedPrefix + headersConfig.DisplayName)
  newUser, err := adminUsers.New(username, "", email, fullname, (s["level"] == adminLevel))
  if err != nil {
    log.Printf("error creating user %s: %v", username, err)
    http.Redirect(w, r, forbiddenPath, http.StatusFound)
    return
  }
} else {
  if err := adminUsers.UpdateMetadata(r.RemoteAddr, r.Header.Get("User-Agent"), username); err != nil {
    log.Printf("error updating metadata for user %s: %v", username, err)
  }
}
```

cmd/admin/auth.go
```diff
@@ -102,8 +103,57 @@ func handlerAuthCheck(h http.Handler) http.Handler {
...
+
+			// This user didn't present a group that has permission to use the service
+			if _, ok := s["level"]; !ok {
+				http.Redirect(w, r, forbiddenPath, http.StatusForbidden)
+				return
+			}
+
+			if !adminUsers.Exists(username) {
+			  log.Printf("User not found, creating: %s", username)
+			  _, err := adminUsers.New(username, "", fullname, (s["level"] == adminLevel))
```
javuto: There is one `string` parameter missing, also my bad since I forgot that `.New()` just creates the `AdminUser` structure, and in order to complete the creation of the user, it needs a subsequent call to `.Create()`, for example:
```
email := r.Header.Get(headersConfig.TrustedPrefix + headersConfig.Email)
newUser, err := adminUsers.New(username, "", email, fullname, (s["level"] == adminLevel))
if err != nil {
  log.Printf("Error with new user %s: %v", username, err)
  http.Redirect(w, r, forbiddenPath, http.StatusFound)
  return
}
if err := adminUsers.Create(newUser); err != nil {
  log.Printf("Error creating user %s: %v", username, err)
  http.Redirect(w, r, forbiddenPath, http.StatusFound)
  return
}
```

obelisk:
Troubleshooting this build error

PR #25, opened 2019-10-21
Fixed Graylog plugin not sending `result`

kosborn:
The objects for type `status` and `result` differ, and `result` was not being converted to JSON.

Duplicated code from the Splunk plugin https://github.com/jmpsec/osctrl/blob/master/plugins/splunk_logging/plugin.go#L42

Validated that it works in GrayLog.


javuto:
Thank you for the PR! 🎉

PR #28, opened 2019-10-27
Adding osctrl-api component

javuto:
## Overview

⚠️ **This is still a Work In Progress PR** ⚠️

* Adding the `osctrl-api` component as a new service. It will listen locally in port `9002` and remotely in `8444`, but those are default values and can be customized.

I will add more details once I am done, since there will be likely changes to this implementation.

cc @obelisk


Vagrantfile
```diff
@@ -3,18 +3,20 @@
 
 VAGRANTFILE_API_VERSION = "2"
 
+IP_ADDRESS = "10.10.10.6"
```
obelisk: Is hardcoding this ok or should we use hostnames (or configurable values)?

cmd/api/db.go
```diff
@@ -0,0 +1,60 @@
...
+	// Performance settings for DB access
+	db.DB().SetMaxIdleConns(dbConfig.MaxIdleConns)
+	db.DB().SetMaxOpenConns(dbConfig.MaxOpenConns)
+	db.DB().SetConnMaxLifetime(time.Second * time.Duration(dbConfig.ConnMaxLifetime))
+
+	return db
+}
+
+// Automigrate of tables
+func automigrateDB() error {
```
obelisk: What is this supposed to do?

cmd/api/handlers-environments.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"log"
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/jmpsec/osctrl/pkg/settings"
+	"github.com/jmpsec/osctrl/pkg/utils"
+)
+
+// GET Handler for single JSON environment
```
obelisk: Do you mean this API is used to validate if an environment exists in the backend?

cmd/api/handlers-environments.go
```diff
@@ -0,0 +1,55 @@
...
+			apiErrorResponse(w, "error getting environment", err)
+		}
+		return
+	}
+	// Header to serve JSON
+	apiHTTPResponse(w, JSONApplicationUTF8, http.StatusOK, env)
+	incMetric(metricAPIOK)
+}
+
+// GET Handler for multiple JSON environments
```
obelisk: And this returns all environments?

cmd/api/handlers-nodes.go
```diff
@@ -0,0 +1,61 @@
...
+		apiErrorResponse(w, "error getting uuid", nil)
+		return
+	}
+	// Get node by UUID
+	node, err := nodesmgr.GetByUUID(uuid)
+	if err != nil {
+		incMetric(metricAPIErr)
+		if err.Error() == "record not found" {
+			log.Printf("node not found: %s", uuid)
+			apiHTTPResponse(w, JSONApplicationUTF8, http.StatusNotFound, ApiErrorResponse{Error: "node not found"})
```
obelisk: Is what this returns a similar JSON format to successful run? If so that'll make it easier to parse in prospective clients.

cmd/api/handlers-nodes.go
```diff
@@ -0,0 +1,61 @@
...
+	apiHTTPResponse(w, JSONApplicationUTF8, http.StatusOK, node)
+	incMetric(metricAPIOK)
+}
+
+// GET Handler for multiple JSON nodes
+func apiNodesHandler(w http.ResponseWriter, r *http.Request) {
+	incMetric(metricAPIReq)
+	utils.DebugHTTPDump(r, settingsmgr.DebugHTTP(settings.ServiceAPI), false)
+	// Get nodes
+	nodes, err := nodesmgr.Gets("all", 0)
```
obelisk: This seems an odd API to use over a more easily locked down `nodesmgr.GetAllNodes()` especially given that `GetAllPlatforms()` exists.

Any reason why you're doing it this way?

cmd/api/handlers-queries.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	// Get query by name
+	query, err := queriesmgr.Get(name)
+	if err != nil {
+		incMetric(metricAPIErr)
+		if err.Error() == "record not found" {
+			log.Printf("query not found: %s", name)
+			apiHTTPResponse(w, JSONApplicationUTF8, http.StatusNotFound, ApiErrorResponse{Error: "query not found"})
+		} else {
+			apiErrorResponse(w, "error getting query", err)
```
obelisk: Can we guarantee this err doesn't contain any sensitive data? It looks like it would be returned to the client (does the client get all that info then)?

cmd/api/handlers-queries.go
```diff
@@ -0,0 +1,174 @@
...
+	incMetric(metricAPIReq)
+	utils.DebugHTTPDump(r, settingsmgr.DebugHTTP(settings.ServiceAPI), false)
+	var q DistributedQueryRequest
+	// Parse request JSON body
+	if err := json.NewDecoder(r.Body).Decode(&q); err != nil {
+		incMetric(metricAPIErr)
+		apiErrorResponse(w, "error parsing POST body", err)
+		return
+	}
+	// FIXME check validity of query
```
obelisk: This is actually quite difficult and might be something you don't want to do because it'll force you to keep in lock step with every device interfacing with osctrl.

What if osquery upgrades SQLite and it adds a new keyword? Now no queries that use that can run through osctrl even though they would normally work. What about different versions of osquery within the same infra, different tables.

I would say not to even worry about this and just make osctrl an extremely reliable middle man :)

cmd/api/handlers-queries.go
```diff
@@ -0,0 +1,174 @@
...
+	// Query can not be empty
+	if q.Query == "" {
+		apiErrorResponse(w, "query can not be empty", nil)
+		return
+	}
+	// Prepare and create new query
+	queryName := "query_" + generateQueryName()
```
obelisk: Is the query prefix needed? If so, that should probably happen inside `generateQueryName()` anyway

cmd/api/handlers-queries.go
```diff
@@ -0,0 +1,174 @@
...
+	newQuery := queries.DistributedQuery{
+		Query:      q.Query,
+		Name:       queryName,
+		Creator:    "API",
```
obelisk: This should be the username from the authorization mechanism, not `API`

cmd/api/handlers-queries.go
```diff
@@ -0,0 +1,174 @@
...
+		Expected:   0,
+		Executions: 0,
+		Active:     true,
+		Completed:  false,
+		Deleted:    false,
+		Repeat:     0,
```
obelisk: I'm not sure how this value should work. In my eyes a distributed query should run only once. If it runs anymore times than that, then when you fetch something by `queryName` what does that actually mean? Is it original data? The most recent data? Both?

cmd/api/handlers-queries.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	// FIXME check validity of query
+	// Query can not be empty
+	if q.Query == "" {
+		apiErrorResponse(w, "query can not be empty", nil)
+		return
+	}
+	// Prepare and create new query
+	queryName := "query_" + generateQueryName()
+	newQuery := queries.DistributedQuery{
```
obelisk: We should have an accelerate flag in here as well. We should set it if a query targets only a single host.

cmd/api/handlers-queries.go
```diff
@@ -0,0 +1,174 @@
...
+		return
+	}
+	// Temporary list of UUIDs to calculate Expected
+	var expected []string
+	// Create environment target
+	if len(q.Environments) > 0 {
+		for _, e := range q.Environments {
+			if (e != "") && envs.Exists(e) {
+				if err := queriesmgr.CreateTarget(queryName, queries.QueryTargetEnvironment, e); err != nil {
+					apiErrorResponse(w, "error creating query environment target", err)
```
obelisk: Same question about sensitive errors getting passed to the client.

cmd/api/handlers-queries.go
```diff
@@ -0,0 +1,174 @@
...
+	if err := queriesmgr.SetExpected(queryName, len(expectedClear)); err != nil {
+		apiErrorResponse(w, "error setting expected", err)
+		return
+	}
+	// Return query name as serialized response
+	apiHTTPResponse(w, JSONApplicationUTF8, http.StatusOK, ApiQueriesResponse{Name: newQuery.Name})
+	incMetric(metricAPIOK)
+}
+
+// GET Handler to return multiple queries in JSON
```
obelisk: This is a super powerful API and should probably be locked to a certain subset of users because that allows visibility into all queries, including ones that might relate to on going investigations.

Vagrantfile
javuto: This is the IP address for the vagrant machine. There is no way (that I know of) to configure this via a command and is better to have the IP set to something known. Since the flow is to run `vagrant up` this should be fine.

cmd/api/db.go
javuto: The `automigrateDB()` function is used by the ORM to create the schema based on the data stored. It is not doing much now, but if we decide to store something in the DB, this will come handy. I am thinking keeping requests per token, so we can enable/enforce some rate limiting?

cmd/api/handlers-environments.go
javuto: Correct, and returns all data for that environment. Does it make more sense to have a verb to check on a specific environment, passing the name, and return just a boolean?

cmd/api/handlers-environments.go
javuto: Yessir.

cmd/api/handlers-nodes.go
javuto: When the requested node exists, it does return the structure for `OsqueryNode{}`, with all its fields in it. Does it make more sense to return always a JSON response with one field for message, and another for data?

cmd/api/handlers-queries.go
javuto: 👍

cmd/api/handlers-queries.go
javuto: Yeah probably, it is also used to generate names for carves, with `carve_` instead of `query_`...

cmd/api/handlers-queries.go
javuto: Totally, but this is just a placeholder until I get the actual tokens part finished.

cmd/api/handlers-queries.go
javuto: This is some legacy stuff, that I thought about having distributed queries in a way that can be repeated, but then again, those queries should be scheduled. I can probably clean this up since it is dead code...

cmd/api/handlers-queries.go
javuto: Let's get that done!

cmd/api/handlers-nodes.go
obelisk: No that should be fine, a client can easily check that the fields are empty :)

PR #35, opened 2020-01-30
small refactor: to use struct for html layout pages

friedbutter:
- Define a new structure to generate the filepaths
- No logic change

Testing...

Reload each page, but if there is another way let me know.


javuto:
Thank you so much for the PR! 🙌 🙏

PR #37, opened 2020-02-20
No more plugins

javuto:
## Overview

Logging data in `osctrl-tls` was using [go plugins](https://golang.org/pkg/plugin/) because whynot, but it turns out that using unstable stuff is tricky so let's keep it in a separate branch and move on.

Now `logging` is its own package and it will contain all supported logging mechanisms.

Also some refactoring to have `backend` as the package that connects to the database.

Thanks in advance for the code review @obelisk @pmoody-


cmd/admin/main.go
```diff
@@ -206,15 +206,21 @@ func init() {
...
+	if err != nil {
+		log.Fatalf("Failed to load DB - %v", err)
+	}
 	// Close when exit
 	//defer db.Close()
 	defer func() {
 		err := db.Close()
 		if err != nil {
-			log.Fatalf("Failed to close Database handler %v", err)
+			log.Fatalf("Failed to close Database handler - %v", err)
```
pmoody-: super nit, 

```
if err := db.Close(); err  != nil {
}
```

pkg/backend/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/jmpsec/osctrl/pkg/backend
```
pmoody-: why the multimodule?

pkg/logging/db.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// LoggerDB will be used to log data using a database
+type LoggerDB struct {
+	Database      *gorm.DB
+	Configuration backend.JSONConfigurationDB
+	Enabled       bool
+}
+
+func CreateLoggerDB() (*LoggerDB, error) {
```
pmoody-: i need to get more used to the github review interface, so I haven't been able to see yet how this is called, but would it make sense to make a Logging interface and have each of files implement a different kind of logger?

pkg/logging/db.go
```diff
@@ -0,0 +1,191 @@
...
+		Enabled:       true,
+	}
+	return l, nil
+}
+
+// Settings - Function to prepare settings for the logger
+func (logDB *LoggerDB) Settings(mgr *settings.Settings) {
+	log.Printf("Setting DB logging settings\n")
+	// Setting link for on-demand queries
+	var _v string
```
pmoody-: nit, 

```
v := settings.QueryLink
```

also, why `_v`?

pkg/logging/db.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+	}
+}
+
+// dbResult - Function that sends JSON result logs to the configured DB
+func (logDB *LoggerDB) Result(data []byte, environment, uuid string, debug bool) {
+	// Parse JSON
+	var logs []types.LogResultData
+	err := json.Unmarshal(data, &logs)
+	if err != nil {
```
pmoody-: nit

```
if err := ...
```

pkg/backend/backend.go
```diff
@@ -0,0 +1,70 @@
+package backend
```
pmoody-: `pkg/backend` seems like an odd path for this rather than something like `backend/`. can you explain this to me tomorrow?

pkg/logging/db.go
```diff
@@ -0,0 +1,191 @@
+package logging
```
pmoody-: same comment here about the `pkg/logging` vs `logging`

pkg/logging/db.go
javuto: Probably that is the way to go

crimsonknave:
@javuto https://osctrl.net/configuration/#logging links to a plugins directory that this PR appears to have removed.

PR #52, opened 2020-03-16
Permissions for users

javuto:
## Overview

* Ability to set permissions for non-admin users from the `osctrl-admin` component.
* Enforcement of permissions for `osctrl-admin`.
* Enforcement of permissions for `osctrl-api`.


admin/handlers-get.go
```diff
@@ -1077,6 +1077,38 @@ func usersGETHandler(w http.ResponseWriter, r *http.Request) {
...
+	// Extract username and verify
+	usernameVar, ok := vars["username"]
+	if !ok || !adminUsers.Exists(usernameVar) {
+		if settingsmgr.DebugService(settings.ServiceAdmin) {
+			log.Printf("DebugService: error getting username")
+		}
+		return
+	}
+	// Get context data
+	ctx := r.Context().Value(contextKey("session")).(contextValue)
```
pmoody-: isn't there a wrapper to pass the contextValue in directly?

admin/handlers-post.go
```diff
@@ -1242,6 +1261,73 @@ send_response:
...
+	// Parse request JSON body
+	if settingsmgr.DebugService(settings.ServiceAdmin) {
+		log.Println("DebugService: Decoding POST body")
+	}
+	if err := json.NewDecoder(r.Body).Decode(&p); err != nil {
+		responseMessage = "error parsing POST body"
+		responseCode = http.StatusInternalServerError
+		if settingsmgr.DebugService(settings.ServiceAdmin) {
+			log.Printf("DebugService: %s %v", responseMessage, err)
+		}
```
pmoody-: return here if err != nil?

admin/handlers-post.go
```diff
@@ -1242,6 +1261,73 @@ send_response:
...
+	} else {
+		// Check CSRF Token
+		if checkCSRFToken(ctx[ctxCSRF], p.CSRFToken) {
```
pmoody-: nit, lose the extra level of indentation

```
if !checkCSRF() {
  // invalid csrf
  return
}
perms := users.UserPermissions{}
```

admin/handlers-post.go
```diff
@@ -1242,6 +1261,73 @@ send_response:
...
+			}
+		} else {
+			responseMessage = "invalid CSRF token"
+			responseCode = http.StatusInternalServerError
+			if settingsmgr.DebugService(settings.ServiceAdmin) {
+				log.Printf("DebugService: %s %v", responseMessage, err)
+			}
+		}
+	}
+send_response:
```
pmoody-: heh, goto.

you could make this a function that gets called everywhere you currently call goto. that'd probably be cleaner. so instead of 

```
				log.Printf("%s has %s", ctx[ctxUser], responseMessage)
				goto send_response
```

you'd have 

```
				log.Printf("%s has %s", ctx[ctxUser], responseMessage)
				sendResponse(<response args>)
                                return
```

admin/templates/users.html
```diff
@@ -119,6 +137,22 @@ <h4 class="modal-title">Add new user</h4>
                         <input class="form-control" name="user_fullname" id="user_fullname" type="text" autocomplete="off">
                       </div>
                     </div>
+                    <div class="form-group row">
+                      <label class="col-md-3 col-form-label" for="user_admin">Enable Admin Level: </label>
+                      <div class="col-md-3">
```
pmoody-: col-md-3 is my favorite div class.

admin/handlers-get.go
javuto: Yes, need to refactor that in eventually

admin/handlers-post.go
javuto: Here it will just go to `send_response:` since it is at the end of the `if`. Again, needs refactor to skip using `goto` and use a function to handle sending responses.

admin/auth.go
```diff
@@ -25,8 +25,54 @@ const (
...
+	return permissions.Carve
+}
+
+// Helper to check if environment access is granted
+func checkEnvironmentLevel(permissions users.UserPermissions, environment string) bool {
+	return permissions.Environments[environment]
+}
+
+// Helper to check permissions for a user
+func checkPermissions(username string, query, carve, env bool, environment string) bool {
```
pmoody-: this is a little bit funky.  if you do

```
ok := getPermissions("peter", true, true, true, "")
```
what actually gets returned is checkQueryLevel, but nothing on checkCarveLevel or checkEnvironmenLevel.

I'd suggest something like

```
type level int
const (
  queryLevel level = iota
  carveLevel
  envLevel
)

func checkPermissions(username string, permissionsLevel level, environment string) bool {
  switch permissionsLevel {
    case queryLevel:
      ..
    case carveLevel:
    ...
  }
```

that would at least make it more obvious that you're only requesting a single permission.

settings/settings.go
```diff
@@ -182,6 +182,19 @@ func (conf *Settings) NewIntegerValue(service, name string, value int64) error {
...
+	validTypes := map[string]bool{
+		TypeString:  true,
+		TypeBoolean: true,
+		TypeInteger: true,
+	}
+	if !validTypes[sType] {
+		return false
+	}
+	return true
+}
```
pmoody-: if you want to do this with a map, i'd make `validTypes` be a global so you're not reconstructing it every time this is called. then you'd check it like

```
var validTypes = map[string]struct{}{
  TypeString: struct{}{},
  TypeBoolean: struct{}{},
  TypeInteger: struct{}{},
}

func (conf *Settings) VerifyType(sType string) bool {
  _, ok := validTypes[sType]
  return ok
}
```

PR #54, opened 2020-03-24
Refactor for TLS handlers

javuto:
## Overview

* Refactor for HTTP handlers, that will allow easier testing.


admin/main.go
```diff
@@ -416,6 +414,9 @@ func main() {
 		}
 	}()
 
+	// multiple listeners channel
+	finish := make(chan bool)
```
pmoody-: where is this used?

logging/dispatch.go
```diff
@@ -0,0 +1,66 @@
...
+	// FIXME allow multiple types of logging
+	if debug {
+		log.Printf("dispatching logs to %s", l.Logging)
+	}
+	l.Log(
+		logType,
+		data,
+		environment,
+		uuid,
+		debug)
```
pmoody-: nit, funny line splitting.

logging/dispatch.go
```diff
@@ -0,0 +1,66 @@
...
+	// Refresh last logging request
+	if logType == types.StatusLog {
+		err := l.Nodes.RefreshLastStatus(uuid)
+		if err != nil {
```
pmoody-:
```
if err := l.Nodes.RefreshLastStatus(uuid); err != nil {
}
```

logging/dispatch.go
```diff
@@ -0,0 +1,66 @@
...
+		log.Printf("dispatching queries to %s", l.Logging)
+	}
+	l.QueryLog(
+		types.QueryLog,
+		data,
+		node.Environment,
+		node.UUID,
+		queryData.Name,
+		queryData.Status,
+		debug)
```
pmoody-: same odd line-splitting comment.

logging/process.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/jmpsec/osctrl/nodes"
+	"github.com/jmpsec/osctrl/types"
+)
+
+// ProcessLogs - Helper to process logs
+func (l *LoggerTLS) ProcessLogs(data json.RawMessage, logType, environment, ipaddress string, debug bool) {
+	// Parse log to extract metadata
+	var logs []types.LogGenericData
+	err := json.Unmarshal(data, &logs)
+	if err != nil {
```
pmoody-:
```
if err := ...; err != nil {
}
```

tls/handlers/utils.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
+
+// Helper to remove duplicates from array of strings
+func uniq(duplicated []string) []string {
+	keys := make(map[string]bool)
+	result := []string{}
+	for _, entry := range duplicated {
+		if _, value := keys[entry]; !value {
+			keys[entry] = true
```
pmoody-: i don't think you actually need this. you don't use it anywhere.

admin/main.go
javuto: Hmmm I had this from a long time ago when I was running all services in the same binary, but now I am not using it anymore. Remove!

PR #324, opened 2022-11-18
Unit test for users

javuto:
Increasing code coverage in the `users` module.


javuto:
```PASS
coverage: 76.2% of statements
ok  	users	0.614s
```

PR #450, opened 2024-07-06
Add support for arm64 docker builds

peterbogdan:
Add support for arm64 docker builds

- Split building images into 2 steps: one to build all images for all platforms and export the digests upstream; create the new image using all digests with `docker buildx imagetools create` as recommended in [multi-platform.md](https://github.com/crazy-max/docker-docs/blob/6aa26f53aeecf40013a3a255b36a8f21a08b0ac9/build/ci/github-actions/multi-platform.md)
- updated `.github/actions/tagged_release/docker/codesign/action.yml` to sign all images and digests pushed to docker hub
- Push `latest` tag on new tagged_releases
- use `docker/metadata-action` to tag the images 
- fixed bug in `.github/actions/tagged_release/github/action.yml` where `inputs.commit_sha` was missing from file name
- fixed bug in `.github/actions/build/dpkg/action.yml` where the artifact could not be found and uploaded

Tagged release workflow tested [here](https://github.com/peterbogdan/osctrl/actions/runs/9825678719)
Normal release workflow tested [here](https://github.com/peterbogdan/osctrl/actions/runs/9825476763)

Tested uploading the new images in my docker hub:
- [osctrl-tls](https://hub.docker.com/r/peterbbb/osctrl-tls/tags)
- [osctrl-cli](https://hub.docker.com/r/peterbbb/osctrl-cli/tags)
- [osctrl-admin](https://hub.docker.com/r/peterbbb/osctrl-admin/tags)
- [osctrl-api](https://hub.docker.com/r/peterbbb/osctrl-api/tags)

### DIFF
--- a/.github/workflows/create_tagged_releases.yml
+++ b/.github/workflows/create_tagged_releases.yml
@@ -84,7 +84,7 @@ jobs:
         shell: bash
         run: |
           echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
-          echo "branch=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_OUTPUT
+          echo "branch=$(echo ${GITHUB_REF_NAME})" >> $GITHUB_OUTPUT
           echo "sha_short=$(git rev-parse --short HEAD)" >> $GITHUB_OUTPUT
 
       ########################### Build DEB packages ###########################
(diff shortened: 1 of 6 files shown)


.github/workflows/create_tagged_releases.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         shell: bash
         run: |
           echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
-          echo "branch=$(echo ${GITHUB_REF#refs/heads/})" >> $GITHUB_OUTPUT
```
peterbogdan, line 87: `GITHUB_REF#refs/heads/` contains `ref/tag` which we don't want if we are going to use it as version tag, not sure if this was the intended way for this var or another one should be used as the input was missing for the action

javuto (approved):

javuto:
Thank you for the pull request! I think at some point I will migrate this process to using `gorelease`, to make it simpler. Nice job! 👏

PR #474, opened 2024-08-08
fix QsqueryQueryData Data type for query results api call

peterbogdan:
Update type for Data as reflected here: https://github.com/jmpsec/osctrl/blob/main/logging/db.go#L54-L54

### DIFF
--- a/api/postgres.go
+++ b/api/postgres.go
@@ -1,8 +1,6 @@
 package main
 
 import (
-	"encoding/json"
-
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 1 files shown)


peterbogdan:
@javuto should this be moved to https://github.com/jmpsec/osctrl/blob/main/types/osquery.go to avoid having to maintain multiple instances of the same type?

javuto:
Hmmm in `osquery.go` we have the standard types for osquery, and the other one is our internal type to handle data serialization. I think it better stays in `db.go` since it is still our internal type.

peterbogdan:
but would it better to be define the type in a single place and use that instead of redefining it in both `logging/db.go` and `api/postgres.go`?

to not keep 2 versions up to date when things are changing

javuto:
Okay that makes sense, and it is better to only have to maintain that type in a single place. Would you add it in this PR? Thank you for working on it! 🙏

peterbogdan:
updated

javuto (approved):

Issue #504, opened 2024-09-11
Supporting prometheus or openmetrics


zhuoyuan-liu:
It would be great if we could support Prometheus or openmetrics to better integrate with other observability tools.

- https://openmetrics.io/
- https://prometheus.io/docs/guides/go-application/
- https://opentelemetry.io/docs/languages/go/

We are happy to contribute to this feature.

javuto:
You have a preference between prometheus or openmetrics? There is an issue to add the `/metrics` endpoint to all services and to use prometheus (#273) but I don't like the model where those metrics need to be exposed. Technically there is nothing sensitive there but still. Thoughts?

zhuoyuan-liu:
All our projects are using Prometheus and it just released their 3.0 last week.

- https://prometheus.io/docs/guides/go-application/
> I don't like the model where those metrics need to be exposed.
This information for sure should not be exposed publicly. It should only be exposed to internal service that can scrape the metrics. We can do it either have a different port for metrics service only or config nginx to only expose required paths.

javuto:
Implemented in https://github.com/jmpsec/osctrl/pull/522 for `osctrl-tls`, more components will follow soon!

PR #522, opened 2024-09-27
Add prometheus metrics to osctrl-tls

zhuoyuan-liu:
This PR added Prometheus metrics support for osctrl-tls. Issue: https://github.com/jmpsec/osctrl/issues/504

Currently, it has all default metrics go metrics such as memory info and gc duration etc. Also, we added one custom histogram metric to measure the total request/latency to each osquery endpoint. 

The metric server would run as a separate service in a different port. We can query e.g. P90 latency for each endpoint with different environments with the histogram metric.

image

Also the request rate for different path and status code. 
image


javuto:
Thank you for the PR! I will add some comments in a code review 🙂

tls/main.go
```diff
@@ -620,6 +630,22 @@ func osctrlService() {
...
 	}()
+
+	// Register Prometheus metrics
+	handlers.RegisterMetrics(prometheus.DefaultRegisterer)
+
+	// Creating a new prometheus service
+	prometheusServer := http.NewServeMux()
+	prometheusServer.Handle("/metrics", promhttp.Handler())
+
+	go func() {
```
javuto, line 641: I think the endpoint `/metrics` should be only enabled if the service is configured to do so, not by default.

types/types.go
```diff
@@ -2,12 +2,13 @@ package types
...
 type JSONConfigurationTLS struct {
-	Listener string `json:"listener"`
-	Port     string `json:"port"`
-	Host     string `json:"host"`
-	Auth     string `json:"auth"`
-	Logger   string `json:"logger"`
-	Carver   string `json:"carver"`
+	Listener    string `json:"listener"`
+	Port        string `json:"port"`
+	MetricsPort string `json:"metricsPort"`
```
javuto, line 7: Adding a new field to enable metrics in the configuration, may be a good idea.

tls/main.go
```diff
@@ -217,6 +219,13 @@ func init() {
 			EnvVars:     []string{"SERVICE_AUTH"},
 			Destination: &tlsConfigValues.Auth,
 		},
+		&cli.StringFlag{
```
javuto, line 222: In order to set the new metrics configuration parameter, it can be done adding a new `BoolFlag` to set it. See the `config-file` or the `logger` flags as examples.

javuto (changes requested):
Thank you for the PR! 🙏 

I have added few comments, mostly related with having metrics not enabled by default, and the ability to enable them via configuration or flags.

zhuoyuan-liu:
I have updated the flag. Should I start to clean up the old metrics components for osctrl-tls?

tls/main.go
```diff
@@ -620,6 +637,24 @@ func osctrlService() {
...
+		// Register Prometheus metrics
+		handlers.RegisterMetrics(prometheus.DefaultRegisterer)
+
+		// Creating a new prometheus service
+		prometheusServer := http.NewServeMux()
+		prometheusServer.Handle("/metrics", promhttp.Handler())
+
+		go func() {
+			log.Info().Msgf("Starting prometheus server at port %s", tlsConfig.MetricsPort)
+			err := http.ListenAndServe(":"+tlsConfig.MetricsPort, prometheusServer)
```
javuto, line 651: I wonder if it makes sense to have an additional flag to provide here, otherwise the listener is going to be `0.0.0.0` by default, without any authentication.

javuto:
> I have updated the flag. Should I start to clean up the old metrics components for osctrl-tls?

You mean the emitted metrics? Instead of removing them, it would be better if we can find a way for those to be utilized by prometheus too.

zhuoyuan-liu:
@javuto I have added the flag, could you please take another look?

javuto (approved):

zhuoyuan-liu:
@javuto I Just fixed the conflicts.

PR #524, opened 2024-10-02
add kafka as a new logging backend

saryani:
This PR will make it possible for osctrl to send the osquery-agent logs to a kafka topic. this means that these logs will no longer be pushed to the database, you need to set kafka as the log backend at init time otherwise it will log into database. 

* added a logger backend to send the agents logs to kafka
* added required configuration definition to allow setting up a kafka logger
* updated the docker-compose to allow to run different backends


javuto:
No problem, the only feedback I had after a quick look, is probably to avoid adding any Kafka related elements to the docker development files, since is not really part of the stack to do local development. Thank you!

Issue #523, opened 2024-09-27
Postgres ssl connection string


zhuoyuan-liu:
At the moment, the SSL connection for Postgres is fixed to be disabled. However, we would like to change it to require. Could we add a parameter to configure this db string?

javuto:
This is a good idea, I have done it here: https://github.com/jmpsec/osctrl/pull/526

PR #526, opened 2024-10-03
Adding flag to utilize the SSL support for the backend connection

javuto:
Using the native SSL support of Postgres to encrypt-decrypt communications to backend. More info at https://www.postgresql.org/docs/current/libpq-ssl.html

Implements #523


zhuoyuan-liu (approved):
LGTM

PR #558, opened 2024-11-05
Refactor distributed query

zhuoyuan-liu:
@javuto , This PR is not completed but shows the idea of how we move all the calculations from each node's distributed query request to the distributed query creation.

In this idea, we would create a new table that records which node should execute which query. In this case, we only iteration all nodes when we create the distributed query. When osquery sends the distributed query result, we only check this table and we don't need to go through the whole list of distributed queries.

I also preferred a small change:
- Use a single `status` column instead of several ones. It would be simplicity and readability. Also, it would be efficient when querying the status, we don't have to write a query for each column.  A classic example would be `"active = ? AND completed = ? AND deleted = ? AND expired = ? AND type = ? AND environment_id = ?",`

In this case, we can get rid of several tables, we only need one table to track the status of all distributed queries.

### DIFF
--- a/queries/queries.go
+++ b/queries/queries.go
@@ -79,6 +79,16 @@ type DistributedQuery struct {
 	Expiration    time.Time
 }
 
+// NodeQuery links a node to a query
+type NodeQuery struct {
+	NodeQueryID uint      `gorm:"primaryKey;autoIncrement"`
+	NodeID      uint      `gorm:"not null;index"`
+	QueryID     uint      `gorm:"not null;index"`
+	Status      string    `gorm:"type:varchar(50);default:'pending'"` // Indexed for fast lookups
+	AssignedAt  time.Time `gorm:"default:CURRENT_TIMESTAMP"`
+	CompletedAt time.Time
+}
+
 // DistributedQueryTarget to keep target logic for queries
 type DistributedQueryTarget struct {
 	gorm.Model
@@ -107,6 +117,7 @@ type Queries struct {
 func CreateQueries(backend *gorm.DB) *Queries {
 	var q *Queries
 	q = &Queries{DB: backend}
+
 	// table distributed_queries
 	if err := backend.AutoMigrate(&DistributedQuery{}); err != nil {
 		log.Fatal().Msgf("Failed to AutoMigrate table (distributed_queries): %v", err)
--- a/queries/queries.go
+++ b/queries/queries.go
@@ -81,12 +81,12 @@ type DistributedQuery struct {
 
 // NodeQuery links a node to a query
 type NodeQuery struct {
-	NodeQueryID uint      `gorm:"primaryKey;autoIncrement"`
-	NodeID      uint      `gorm:"not null;index"`
-	QueryID     uint      `gorm:"not null;index"`
-	Status      string    `gorm:"type:varchar(50);default:'pending'"` // Indexed for fast lookups
-	AssignedAt  time.Time `gorm:"default:CURRENT_TIMESTAMP"`
-	CompletedAt time.Time
+	ID        uint      `gorm:"primaryKey;autoIncrement"`
+	NodeID    uint      `gorm:"not null;index"`
+	QueryID   uint      `gorm:"not null;index"`
+	Status    string    `gorm:"type:varchar(10);default:'pending'"`
+	CreatedAt time.Time `gorm:"default:CURRENT_TIMESTAMP"`
+	UpdatedAt time.Time
 }
 
 // DistributedQueryTarget to keep target logic for queries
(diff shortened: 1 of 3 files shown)


queries/queries.go
```diff
@@ -79,6 +79,16 @@ type DistributedQuery struct {
 	Expiration    time.Time
 }
 
+// NodeQuery links a node to a query
+type NodeQuery struct {
+	ID        uint      `gorm:"primaryKey;autoIncrement"`
+	NodeID    uint      `gorm:"not null;index"`
+	QueryID   uint      `gorm:"not null;index"`
+	Status    string    `gorm:"type:varchar(10);default:'pending'"`
```
zhuoyuan-liu, line 87: We can also use ENUM type here for more consistency

`CREATE TYPE task_status AS ENUM ('pending', 'completed', 'error');`

zhuoyuan-liu:
@javuto Could you please take a look and give some feedback? It was tested in our dev cluster. All the core functions work well. 

I tried to make this PR as small as possible and only replaced the way of creating new queries.

Here are the follow-up tasks after this PR:
Update the logic of completing a distributed query
Update the logic showing expected node and completed node
Remove several tables that are no longer needed

admin/handlers/post.go
```diff
@@ -223,6 +229,18 @@ func (h *HandlersAdmin) QueryRunPOSTHandler(w http.ResponseWriter, r *http.Reque
...
 	expectedClear := removeStringDuplicates(expected)
+
+	// Create new record for query list
+	for _, nodeUUID := range expectedClear {
+		node, err := h.Nodes.GetByUUID(nodeUUID)
+		if err != nil {
+			log.Err(err).Msgf("error getting node %s and failed to create node query for it", nodeUUID)
+			continue
+		}
+		if err := h.Queries.CreateNodeQuery(node.ID, newQuery.ID); err != nil {
```
zhuoyuan-liu, line 240: Currently, it creates a single record every time for each node. I think it could be optimized by using batch creation.

queries/queries.go
```diff
@@ -79,6 +80,16 @@ type DistributedQuery struct {
 	Expiration    time.Time
 }
 
+// NodeQuery links a node to a query
+type NodeQuery struct {
```
javuto, line 84: Take a look at other tables defined by gorm, you can add here the `gorm.Model` which will automatically add to the table the following columns:
```
    ID uint `gorm:"primarykey"`
    CreatedAt time.Time
    UpdatedAt time.Time
    DeletedAt DeletedAt `gorm:"index"`
```

queries/queries.go
javuto: The ORM I am using (https://gorm.io/) supports ENUMs, so we can explore using that: https://techwasti.com/how-to-use-enums-with-gorm

queries/queries.go
```diff
@@ -126,34 +142,29 @@ func CreateQueries(backend *gorm.DB) *Queries {
 	return q
 }
 
-// NodeQueries to get all queries that belong to the provided node
-// FIXME this will impact the performance of the TLS endpoint due to being CPU and I/O hungry
-// FIMXE potential mitigation can be add a cache (Redis?) layer to store queries per node_key
```
javuto, line 131: Nit: typo FIXME

queries/queries.go
```diff
@@ -449,6 +472,35 @@ func (q *Queries) SetExpected(name string, expected int, envid uint) error {
 	return nil
 }
 
+// UpdateQueryStatus to update the status of each query
+func (q *Queries) UpdateQueryStatus(queryName string, nodeID uint, statusCode int) error {
+
+	var result string
+	if statusCode == 0 {
+		result = "completed" // TODO: need be replaced with a constant
```
javuto, line 480: I agree we should use `const`

queries/queries_test.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	nodeQuery := queries.NodeQuery{
+		NodeID:  1,
+		QueryID: 1,
+		Status:  "pending",
+	}
+
+	// Query sqlite_master to list all tables
+	var tables []string
+	err = db.Raw("SELECT name FROM sqlite_master WHERE type='table'").Scan(&tables).Error
```
javuto, line 53: The backend is postgres, so as long as we don't use anything postgres specific, we can still test the code using sqlite. Otherwise, testing it is a pain due to all the mocking, see [users/users_test.go](https://github.com/jmpsec/osctrl/blob/main/users/users_test.go)

zhuoyuan-liu: Yes, I agree, we can use varchar for now for easy testing.

queries/queries.go
zhuoyuan-liu: Just realize that SQLite does not support ENUM and we can use the char for now and update later if we see another bottleneck issue.

javuto (approved):

PR #567, opened 2024-11-12
Add log level and format configuration options for services

zhuoyuan-liu:
Introduce configuration options for log level and format in services, allowing customization of logging behavior. This change enhances the flexibility of logging by enabling different log levels and formats through command-line flags and environment variables.

The change was made for admin and api. For tls, I would like to discuss a better naming convention:
- We need a configuration for server's logs
- We need another logger configuration for exports logs and results from osquery

Also, the code are highly duplicated and I would like to remove the duplication.

For documentation, do you know where should update these configuration options?

### DIFF
--- a/admin/main.go
+++ b/admin/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/crewjam/saml/samlsp"
@@ -256,6 +257,20 @@ func init() {
 			Usage:       "TCP port for the service",
 			EnvVars:     []string{"SERVICE_PORT"},
 			Destination: &adminConfigValues.Port,
+		},
+		&cli.StringFlag{
+			Name:        "log-level",
+			Value:       "info",
+			Usage:       "Log level for the service",
+			EnvVars:     []string{"SERVICE_LOG_LEVEL"},
+			Destination: &adminConfigValues.LogLevel,
+		},
+		&cli.StringFlag{
+			Name:        "log-format",
+			Value:       "json",
+			Usage:       "Log format for the service",
+			EnvVars:     []string{"SERVICE_LOG_FORMAT"},
+			Destination: &adminConfigValues.LogFormat,
 		},
 		&cli.StringFlag{
 			Name:        "auth",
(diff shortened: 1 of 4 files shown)


javuto:
This is great, thank you for working on it!

If you are thinking having different flags for `osctrl-tls`, it would be good to follow the same convention for the three services. 

What do you mean with `We need another logger configuration for exports logs and results from osquery`?

To update the documentation, clone the repository https://github.com/jmpsec/osctrl-docs, update the markdown in the `content` folder, and just run `make` to generate new HTML docs. Once the PR is merged, the documentation at https://osctrl.net will be automatically updated.

admin/main.go
```diff
@@ -1046,8 +1085,13 @@ func main() {
 	}
 	app.Action = cliAction
 	if err := app.Run(os.Args); err != nil {
-		log.Fatal().Msgf("app.Run error: %v", err)
```
javuto, line 1049: It is preferable to use `Fatal()` than to print the error and call exit, because closing writers and flushing buffers. See the implementation of `Fatal()` in `zerolog`:
```
// Fatal starts a new message with fatal level. The os.Exit(1) function
// is called by the Msg method, which terminates the program immediately.
//
// You must call Msg on the returned event in order to send the event.
func (l *Logger) Fatal() *Event {
	return l.newEvent(FatalLevel, func(msg string) {
		if closer, ok := l.w.(io.Closer); ok {
			// Close the writer to flush any buffered message. Otherwise the message
			// will be lost as os.Exit() terminates the program immediately.
			closer.Close()
		}
		os.Exit(1)
	})
}```

File: admin/main.go [hunk 2]
```diff
@@ -1026,6 +1036,35 @@ func cliAction(c *cli.Context) error {
 	return nil
 }
 
+func initializeLogger(logLevel, logFormat string) {
```
javuto, line 1039: The strings can be `const`

admin/main.go
```diff
@@ -257,6 +258,20 @@ func init() {
 			EnvVars:     []string{"SERVICE_PORT"},
 			Destination: &adminConfigValues.Port,
 		},
+		&cli.StringFlag{
+			Name:        "log-level",
+			Value:       "info",
```
javuto, line 263: Default value can be a `const`, similarly to the others https://github.com/jmpsec/osctrl/blob/main/admin/main.go#L35

admin/main.go
```diff
@@ -257,6 +258,20 @@ func init() {
...
+			Usage:       "Log level for the service",
+			EnvVars:     []string{"SERVICE_LOG_LEVEL"},
+			Destination: &adminConfigValues.LogLevel,
+		},
+		&cli.StringFlag{
+			Name:        "log-format",
+			Value:       "json",
```
javuto, line 270: Ditto

api/main.go
```diff
@@ -204,6 +205,20 @@ func init() {
 			EnvVars:     []string{"SERVICE_PORT"},
 			Destination: &apiConfigValues.Port,
 		},
+		&cli.StringFlag{
+			Name:        "log-level",
+			Value:       "info",
```
javuto, line 210: Same as `admin/main.go`

javuto (changes requested):
I think we can use the same parameters in `osctrl-tls` for consistency, but there are some changes needed, thank you!

zhuoyuan-liu (commented):
@javuto  I have updated the code according to your comments, could you please take another look.

PR #576, opened 2024-12-16
Second step for refactor distributed query

zhuoyuan-liu:
In this step, we would like to refactor the way of completing a distributed query and minimize the write request to the database.

Changelog:
- Removed TrackExcution function and relevant table
- Replaced `VerifyComplete` with `CleanupCompletedQueries`. In this case, we don't need to check the complete status whenever we receive new data from Osquery. Instead, we check it periodically with the expired function. Since we changed how to fetch the list of target queries, keeping a distributed query in an active state does not affect performance.
- Update how to get lists of target nodes in API and osctrl-admin. We use the intersection to get a list of targeted nodes. We can easily extend this part to support more tags and unions or other operations.

TODO:
- Remove everything related to `CreateTarget` function and relevant database table. This function was used by admin, API and cli. I am wondering if we can remove these duplications and make sure the Admin and CLI talk to API instead of the database directly.


javuto (approved):
Nice work! 👏

Issue #577, opened 2024-12-27
On-demand queries never get completed


javuto:
If an on-demand query is queued up, and the target picks it up and results are returned. But the results keep coming in. It may be related to what merged in https://github.com/jmpsec/osctrl/pull/558

javuto:
Found the root cause:

```
osctrl-tls-dev       | 2024/12/27 19:31:40 /usr/src/app/queries/queries.go:502 ERROR: value too long for type character varying(8) (SQLSTATE 22001)
osctrl-tls-dev       | [0.466ms] [rows:0] UPDATE "node_queries" SET "status"='completed',"updated_at"='2024-12-27 19:31:40.977' WHERE "node_queries"."deleted_at" IS NULL AND "id" = 2
osctrl-postgres-dev  | 2024-12-27 19:31:40.982 UTC [940] ERROR:  value too long for type character varying(8)
osctrl-tls-dev       | {"level":"error","error":"ERROR: value too long for type character varying(8) (SQLSTATE 22001)","time":"2024-12-27T19:31:40Z","caller":"/usr/src/app/logging/process.go:91","message":"error updating query status"}
osctrl-postgres-dev  | 2024-12-27 19:31:40.982 UTC [940] STATEMENT:  UPDATE "node_queries" SET "status"=$1,"updated_at"=$2 WHERE "node_queries"."deleted_at" IS NULL AND "id" = $3
```

The `varchar(8)` introduced in #558 can not fit the word `completed` so I will change it to `varchar(10)` to have on-demand queries operational again.

javuto:
Confirmed fix works on a fresh environment: https://github.com/jmpsec/osctrl/pull/578

PR #578, opened 2024-12-27
Fix to prevent queries not completing

javuto:
Fix for #577 

Schema for the table `node_queries` must be altered manually to expand its size. Use this query:

```
ALTER TABLE node_queries ALTER COLUMN status TYPE varchar (10);
```

This should resolve the issue.


zhuoyuan-liu (approved):
LGTM, sorry for introducing this bug.

PR #593, opened 2025-02-08
Adding types to tags

javuto:
Adding the ability to classify tags with types, depending if they are an environment, platform, localname, uuid or a custom value. This will help to group nodes based on the tags and utilize them for things like on-demand queries, carves and more.


javuto:
In order to populate the `tag_type` field, use a query like the following:

```
UPDATE admin_tags
SET tag_type = CASE admin_tags.name
    WHEN osquery_nodes.environment THEN 0
    WHEN osquery_nodes.uuid THEN 1
    WHEN osquery_nodes.platform THEN 2
    WHEN osquery_nodes.localname THEN 3
    ELSE 4
END
FROM osquery_nodes
WHERE admin_tags.environment_id = osquery_nodes.environment_id;
```

PR #607, opened 2025-02-28
Refactor and enhance pipeline with batch processing

zhuoyuan-liu:
This update introduces a new batch processing mechanism for node last seen time udpate, improving efficiency and reducing redundant operations.

It also removed unused functions. Discontinued historical tracking of IP addresses and node names, as there are no clear use cases for this data.

What is missing in this PR:

- Graceful Shutdown for the Main Function: The main function needs to be updated to allow graceful shutdown to ensure cached data is properly flushed before shutdown.


pkg/nodes/nodes.go
```diff
@@ -585,19 +459,9 @@ func (n *NodeManager) IncreaseBytes(node OsqueryNode, incBytes int) error {
...
-	if lastIp != "" {
-		updates["ip_address"] = lastIp
-	}
-	if err := n.DB.Model(&node).Updates(updates).Error; err != nil {
-		return fmt.Errorf("Updates %v", err)
-	}
-	return nil
+func (n *NodeManager) RefreshLastSeenBatch(nodeID []uint) error {
+
+	return n.DB.Model(&OsqueryNode{}).Where("id IN ?", nodeID).UpdateColumn("last_config", time.Now()).Error
```
javuto, line 464: I think we can just have a new field called `last_seen` and not to recycle any of the existing ones.

cmd/tls/handlers/writers.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"time"
+
+	"github.com/jmpsec/osctrl/pkg/nodes"
+	"github.com/rs/zerolog/log"
+)
+
+// writeEvent represents a single update request.
+type writeEvent struct {
```
javuto, line 11: Either we have a typed event for each update, or something generic enough that can be used for any updates

javuto (changes requested):
See the two comments about having a new field `last_seen` and to have generic events. Thanks for the PR!

cmd/tls/handlers/writers.go
zhuoyuan-liu: For now, the only thing we need to refresh is the `last_seen` time. Do you have any idea for additional resources we need to write/refresh frequently?

cmd/tls/handlers/handlers.go
```diff
@@ -180,12 +181,21 @@ func WithLogs(logs *logging.LoggerTLS) Option {
...
+	}
+}
+
 // CreateHandlersTLS to initialize the TLS handlers struct
 func CreateHandlersTLS(opts ...Option) *HandlersTLS {
 	h := &HandlersTLS{}
 	for _, opt := range opts {
 		opt(h)
 	}
+	// All these opt function need be refactored to reduce unnecessary complexity
```
zhuoyuan-liu, line 197: I would like to refactor the whole main.go to reduce the complexity of loading the application.

cmd/tls/handlers/writers.go
```diff
@@ -0,0 +1,93 @@
...
+
+// flush performs the bulk update for a batch of events.
+func (bw *batchWriter) flush(batch map[uint]writeEvent) {
+
+	nodeIDs := make([]uint, 0, len(batch))
+	for _, ev := range batch {
+		nodeIDs = append(nodeIDs, ev.NodeID)
+
+		// TODO: Implement the actual update logic.
+		// Update the node's IP address.
```
zhuoyuan-liu, line 86: I am wondering if we need to record the IP address of the devices. If the users are under a proxy or VPN, we always get the incorrect IP address.

zhuoyuan-liu (commented):
@javuto  thanks for the comments. I have updated the code, could you please take another look?

cmd/tls/handlers/writers.go
javuto: I think having only the `last_seen` time, updating the `last_ip_address` and keeping the metadata refreshed with the decorator queries should be the only fields that need to be refreshed. Another option would be to implement functionality that runs on-demand queries when certain information needs to be updated.

cmd/tls/handlers/writers.go
javuto: I think is some information useful in the event of an incident. Even if it is just the proxy or VPN, that also provides signal useful for incident responders.

javuto (approved):

pkg/nodes/nodes.go
```diff
@@ -457,10 +457,11 @@ func (n *NodeManager) RefreshLastSeenBatch(nodeID []uint) error {
 	return n.DB.Model(&OsqueryNode{}).Where("id IN ?", nodeID).UpdateColumn("last_seen", time.Now()).Error
```
javuto, line 457: This can be `id = ?` if it is just a single `nodeID`

zhuoyuan-liu: It will update the time in batch. So, it has multiple IDs. I think I will update the main.go in the next step and change this updating events to an interface so that we can decide to either sen data to redis and hand the write in another place or just use the channel.

cmd/tls/handlers/utils.go
```diff
@@ -95,11 +95,7 @@ func nodeFromEnroll(req types.EnrollRequest, env environments.TLSEnvironment, ip
 		ConfigHash:      req.HostDetails.EnrollOsqueryInfo.ConfigHash,
 		BytesReceived:   recBytes,
 		RawEnrollment:   string(enrollRaw),
-		LastStatus:      time.Time{},
-		LastResult:      time.Time{},
-		LastConfig:      time.Time{},
-		LastQueryRead:   time.Time{},
-		LastQueryWrite:  time.Time{},
```
apollo13, line 102: The node.html Template also needs updating now I guess?

javuto: Good catch, I completely missed that. Let me prepare a quick PR to make sure `osctrl-admin` does not break. Thanks!

PR #617, opened 2025-03-13
Use the `last_seen` field in the `node.html` template

javuto:
With the changes landed in #607 some of the node columns were removed and a new one was added to keep when the node was last seen. This PR fixes the HTML template for `osctrl-admin` to prevent accessing the removed fields.
Also updated the OpenAPI yaml.


zhuoyuan-liu (approved):
LGTM

PR #618, opened 2025-03-16
Last refactor changes from `golangci-lint run`

javuto:
* Last remaining changes from running `golangci-lint run`
* Removed `gofmt` from `Makefile` and added `make lint` with the `golangci-lint run` command


zhuoyuan-liu (approved):
LGTM, shall we add the golangci-lint to the github action check as well?

javuto:
Yes I think it makes sense for PRs to run it always and keep the code sane. Thank you!

PR #619, opened 2025-03-18
Move cli flags to a separate pkg 

zhuoyuan-liu:
In this PR, we moved the cli flags into a separate package and therefore these flags can be reused among different components.


pkg/config/flags.go
```diff
@@ -0,0 +1,552 @@
...
+	// Logger flags
+	LoggerFile   string
+	LoggerDbSame bool
+	AlwaysLog    bool
+
+	// Carver flags
+	CarverConfigFile string
+
+	// Configuration values
+	TLSConfigValues    types.JSONConfigurationTLS
```
zhuoyuan-liu, line 37: these configuration files are in `types` pkg and it would be more reasonable to move to this config pkg. e.g. under `config.go` files.

pkg/config/flags.go
```diff
@@ -0,0 +1,552 @@
...
+	TLSWriterConfig    types.JSONConfigurationTLSWriter
+	DBConfigValues     backend.JSONConfigurationDB
+	RedisConfigValues  cache.JSONConfigurationRedis
+	S3LogConfig        types.S3Configuration
+	S3CarverConfig     types.S3Configuration
+	KafkaConfiguration types.KafkaConfiguration
+}
+
+// InitTLSFlags initializes all the flags needed for the TLS service
+func InitTLSFlags(params *TLSFlagParams) []cli.Flag {
```
zhuoyuan-liu, line 47: For other components, it can just pick the required flags instead of copying whole flags.

javuto (approved):

pkg/config/flags.go
javuto: I can migrate all those values to the new config pkg

PR #622, opened 2025-03-19
Migrate all configuration consts from types to config pkg

javuto:
Follow up refactor from the newly added pkg `config` in #619 , to migrate all the configuration related consts from the `types` pkg to `config`. It impacts a lot of other pkgs and all services.


go.mod
```diff
@@ -69,6 +75,7 @@ require (
 	github.com/jinzhu/inflection v1.0.0 // indirect
 	github.com/jinzhu/now v1.1.5 // indirect
 	github.com/jmespath/go-jmespath v0.4.0 // indirect
+	github.com/jmpsec/osctrl/types v0.0.0-20250216225221-94ec48a2c3fe
```
zhuoyuan-liu, line 78: I think some of the packages or references are not fully updated. Otherwise, we should not require this package as they share the same go.mod file.

javuto: It looks good after running `go mod tidy`, https://github.com/jmpsec/osctrl/pull/624

PR #625, opened 2025-03-20
Fix for `osctrl-tls` after refactoring flags and configuration

javuto:
Fix for `osctrl-tls` after the flags refactor into its own `config` pkg in #619 

With this PR the docker dev environment is operational again.


zhuoyuan-liu (approved):
Good job, It's a further refactor to avoid unnecessary variables. However, do you find the root cause for the db flags missing?

PR #631, opened 2025-03-27
Fix error when creating node queries for large datasets

zhuoyuan-liu:
Implement batch creation of node queries to handle more than 10,000 nodes efficiently.

It's a quick fix and I think it's a good idea to applied the batchsize as a global configuration for DB cofig.

It can be done by:
```go
db, err := gorm.Open(postgres.Open(dsn), &gorm.Config{
    CreateBatchSize: 1000, 
})
```

### DIFF
--- a/pkg/queries/queries.go
+++ b/pkg/queries/queries.go
@@ -408,7 +408,7 @@ func (q *Queries) CreateNodeQueries(nodeIDs []uint, queryID uint) error {
 			QueryID: queryID,
 		})
 	}
-	if err := q.DB.Create(&nodeQueries).Error; err != nil {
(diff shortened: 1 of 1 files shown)


javuto:
Is there something missing in this PR?

javuto (approved):

javuto:
Nevermind, I found the documentation https://gorm.io/docs/create.html#Batch-Insert

javuto:
If we apply the configuration using the `gorm.Config` statement, then every `Create` will be affected by that batch size. Not sure if that could have a negative impact or not.

zhuoyuan-liu:
Here is the source code:
image
image

For a single value creation, it functions the same as a regular create operation. As for the rest of the database operations, I don't see any batch creation involved, so this change should have no impact on other operations.

PR #632, opened 2025-03-28
Refactor calls to debug messages in all services

javuto:
- Refactor in `osctrl-tls`, `osctrl-admin` and `osctrl-api` to all calls of debug messages to avoid getting the setting value from the backend.
- Added three new flags to all services to set debug HTTP settings.
* `--enable-http-debug`
* `--http-debug-file`
* `--http-debug-show-body`
- When enabled, the debug HTTP logs will go into a separate file.

### DIFF
--- a/cmd/api/handlers/handlers.go
+++ b/cmd/api/handlers/handlers.go
@@ -5,12 +5,14 @@ import (
 	"github.com/jmpsec/osctrl/pkg/carves"
 	"github.com/jmpsec/osctrl/pkg/config"
 	"github.com/jmpsec/osctrl/pkg/environments"
+	"github.com/jmpsec/osctrl/pkg/logging"
 	"github.com/jmpsec/osctrl/pkg/nodes"
 	"github.com/jmpsec/osctrl/pkg/queries"
 	"github.com/jmpsec/osctrl/pkg/settings"
 	"github.com/jmpsec/osctrl/pkg/tags"
 	"github.com/jmpsec/osctrl/pkg/users"
 	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
 	"gorm.io/gorm"
 )
 
@@ -102,10 +104,24 @@ func WithName(name string) HandlersOption {
 	}
 }
 
-func WithDebugHTTP(logger *zerolog.Logger, cfg *config.DebugHTTPConfiguration) HandlersOption {
+func WithDebugHTTP(cfg *config.DebugHTTPConfiguration) HandlersOption {
 	return func(h *HandlersApi) {
-		h.DebugHTTP = logger
 		h.DebugHTTPConfig = cfg
+		h.DebugHTTP = nil
+		if cfg.Enabled {
+			l, err := logging.CreateDebugHTTP(cfg.File, logging.LumberjackConfig{
+				MaxSize:    25,
+				MaxBackups: 5,
+				MaxAge:     10,
+				Compress:   true,
+			})
+			if err != nil {
+				log.Err(err).Msg("error creating debug HTTP logger")
+				l = nil
+				h.DebugHTTPConfig.Enabled = false
+			}
+			h.DebugHTTP = l
+		}
 	}
 }
 
--- a/cmd/api/handlers/handlers.go
+++ b/cmd/api/handlers/handlers.go
@@ -10,6 +10,7 @@ import (
 	"github.com/jmpsec/osctrl/pkg/settings"
 	"github.com/jmpsec/osctrl/pkg/tags"
 	"github.com/jmpsec/osctrl/pkg/users"
+	"github.com/rs/zerolog"
 	"gorm.io/gorm"
 )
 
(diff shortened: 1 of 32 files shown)


javuto:
@zhuoyuan-liu please take a look, thanks!

zhuoyuan-liu (commented):
Good job! I left a small comment. However, I do have some suggestions if you're interested in further modernizing osctrl.

1. Reducing Global Variables and Direct Dependencies
Relying too much on global variables and direct dependencies can make the code messy and hard to manage. It can lead to unexpected bugs, tricky testing, and parts of the code depending on each other in ways that aren’t obvious. Using dependency injection helps by making it clear what each part of the code needs, making testing easier, and allowing components to be reused more easily.

2. Separating Handler Responsibilities
When handlers handle too much—like HTTP requests, business logic, and database work—it leads to messy and repetitive code. Splitting things into layers, where handlers only deal with HTTP, services handle logic, and repositories manage data, makes the code easier to test, reuse, and maintain. It also keeps errors and responses consistent across the application.

pkg/utils/http-utils.go
```diff
@@ -125,28 +125,24 @@ func SendRequest(reqType, reqURL string, params io.Reader, headers map[string]st
...
 	return debug
 }
 
 // DebugHTTPDump - Helper for debugging purposes and dump a full HTTP request
-func DebugHTTPDump(r *http.Request, debugCheck bool, showBody bool) {
-	if debugCheck {
-		log.Debug().Msg(DebugHTTP(r, debugCheck, showBody))
-	}
+func DebugHTTPDump(r *http.Request, showBody bool) {
+	log.Debug().Msg(DebugHTTP(r, showBody))
```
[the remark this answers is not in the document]
javuto: Just pushed these changes, let me know what you think! And if you have more examples of how to implement the two points mentioned above, feel free to point them out, I am always happy to improve osctrl!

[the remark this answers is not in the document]
zhuoyuan-liu: I would suggest putting the check in this function to make the code more readable.

```go
func DebugHTTPDump(r *http.Request, showBody bool) {
    if log.Debug().Enabled() {
	log.Debug().Msg(DebugHTTP(r, showBody))
    }
}
```

pkg/utils/http-utils.go
```diff
@@ -125,28 +125,24 @@ func SendRequest(reqType, reqURL string, params io.Reader, headers map[string]st
...
-		if err != nil {
-			log.Err(err).Msg("error while dumprequest")
-		}
-		debug += fmt.Sprintf("%s\n", string(requestDump))
-		if !showBody {
-			debug += fmt.Sprintf("%s\n", "---------------- No Body")
-		}
-		debug += fmt.Sprintf("%s\n", "---------------- end")
+	debug = fmt.Sprintf("%s\n", "---------------- request")
+	requestDump, err := httputil.DumpRequest(r, showBody)
```
zhuoyuan-liu, line 131: I just realized that it logs everything from the request, which seems excessive and comes with several drawbacks:

1. **Log Volume** – It generates a large number of logs, making search and analysis more difficult.
2. **Security Concerns** – Some headers may contain sensitive information that should not be logged.
3. **Formatting** – I'm unsure how this would look when integrated with `zerolog`. Do you have an example?

In production, these logs might add too much noise, making it harder to diagnose issues when something goes wrong. I’d love to hear more about how you use these logs and whether there are ways to refine them.

javuto: This is a good point, I believe this was a function I used a lot in the early stages of the osctrl development, but now it would make more sense to:

zhuoyuan-liu: If these logs have been rarely used recently, I’d suggest removing them. In our use case, osctrl handles 100+ requests per second, and these logs would likely add unnecessary pressure.

I understand they might be useful in certain cases, especially during development. If we need to keep them, we could use a feature flag to control logging, enabling it only in the development environment. So, it would not be mixed with other debug log messages.

E.g. --enable-http-dump=true

javuto: I just added a new commit with three new flags in all services:

- `--enable-http-debug`
- `--http-debug-file`
- `--http-debug-show-body`

So if needed, this must be enabled with a flag, which will reduce the risk of logs taking up too much space. And the logs will go to a different file and not stdout.

Let me know what you think! Thanks!

cmd/api/handlers/handlers.go
```diff
@@ -99,6 +102,13 @@ func WithName(name string) HandlersOption {
 	}
 }
 
+func WithDebugHTTP(logger *zerolog.Logger, cfg *config.DebugHTTPConfiguration) HandlersOption {
+	return func(h *HandlersApi) {
+		h.DebugHTTP = logger
```
zhuoyuan-liu, line 107: It could be a good example of "Reducing Global Variables and Direct Dependencies". Instead of relying on the global log variable/package, you can assign the logger to your main server(strcut).

cmd/tls/main.go
```diff
@@ -390,18 +393,33 @@ func initializeLogger(logLevel, logFormat string) {
...
 		}
-		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: "2006-01-02T15:04:05.999Z07:00"}).With().Caller().Logger()
+		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: loggerTimeFormat}).With().Caller().Logger()
 	default:
 		log.Logger = log.With().Caller().Logger()
 	}
+	// If enabled, prepare debug HTTP logger
+	if debugHTTP.Enabled {
+		// Open or create the debug HTTP log file (append mode)
+		debugFile, err := os.OpenFile(
```
zhuoyuan-liu, line 411: Do you have any idea about the log file truncate or rotation? If it is deployed in k8s cluster or docker, you may need extra tools/steps for handling these log files. Since we would never enable them in Prod, it should be fine.

zhuoyuan-liu (commented):
LGTM. However, I believe that as osctrl matures, we will eventually remove these debug logs since they reduce code readability and provide limited value.

PR #636, opened 2025-04-03
Add inmemory cache for env

zhuoyuan-liu:
This PR introduces an in-memory cache to store parameters that rarely change, reducing the number of requests to the backend database. However, it should only be used for bounded values—such as the environment and node key—because the cache is currently unbounded and may grow indefinitely, potentially leading to OOMKilled errors.

At this point, the cache has been added for the `environment` value only.

### DIFF
--- a/pkg/environments/env-cache.go
+++ b/pkg/environments/env-cache.go
@@ -7,6 +7,10 @@ import (
 	"github.com/jmpsec/osctrl/pkg/cache"
 )
 
+const (
+	cacheName = "environments"
+)
+
 // EnvCache provides cached access to TLS environments
 type EnvCache struct {
 	// The cache itself, storing Environment objects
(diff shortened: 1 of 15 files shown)


pkg/environments/environments.go
```diff
@@ -109,14 +109,14 @@ type MapEnvByID map[uint]NameUUID
 // MapEnvByString to hold the environments name and UUID by string
 type MapEnvByString map[string]NameUUID
 
-// Environment keeps all TLS Environments
-type Environment struct {
+// EnvironmentManager keeps all TLS Environments
+type EnvironmentManager struct {
```
javuto, line 113: nit: Maybe the name can be `EnvManager` to reduce its length

pkg/environments/env-cache.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Reference to the environment manager for cache misses
+	envs EnvManager
+}
+
+// NewEnvCache creates a new environment cache
+func NewEnvCache(envs EnvManager) *EnvCache {
+	// Create a new cache with a 10-minute cleanup interval
+	envCache := cache.NewMemoryCache(
+		cache.WithCleanupInterval[TLSEnvironment](2 * time.Hour),
```
zhuoyuan-liu, line 23: This value is currently hardcoded, and I think we need a more robust approach to configuration management. Otherwise, the number of command-line flags will quickly become unmanageable and difficult to maintain.

javuto (approved):
Nicely done!

PR #640, opened 2025-04-08
Refactor nodes package

zhuoyuan-liu:
I’m in the middle of adding in-memory caching to the node packages, and I’ve noticed quite a few areas that could be improved along the way.

This also includes a bug fix related to how active nodes are identified. I would expect a longer time for this PR.

---- 

There is an import cycle:
```
pkg/nodes/node-cache.go imports pkg/cache
pkg/cache → pkg/types → pkg/queries → pkg/nodes
```
My idea is that the cache package should be generic and not import any other packages. The same applies to the types package, as it only stores some of the osquery data formats.


pkg/nodes/nodes.go
```diff
@@ -204,16 +130,16 @@ func (n *NodeManager) GetBySelector(stype, selector, target string, hours int64)
 			return nodes, err
 		}
 	case ActiveNodes:
-		// if err := n.DB.Where(s+" = ?", selector).Where("updated_at > ?", time.Now().AddDate(0, 0, -3)).Find(&nodes).Error; err != nil {
-		if err := n.DB.Where(s+" = ?", selector).Where("updated_at > ?", time.Now().Add(time.Duration(hours)*time.Hour)).Find(&nodes).Error; err != nil {
+		if err := n.DB.Where(s+" = ?", selector).Where("last_seen > ?", time.Now().Add(time.Duration(hours)*time.Hour)).Find(&nodes).Error; err != nil {
```
javuto, line 133: The field `updated_at` should be refreshed every time there is an `UPDATE` operation. Have you noticed a significant discrepancy in those times?

zhuoyuan-liu: I think it's because we are using the `UpdateColumn` for updating `last_seen`, which will update only specific columns and skip the model hooks to improve the performance.

pkg/nodes/nodes.go
```diff
@@ -128,31 +66,32 @@ func (n *NodeManager) CheckByUUIDEnv(uuid, environment string) bool {
...
 	var node OsqueryNode
 	if err := n.DB.Where("node_key = ?", strings.ToLower(nodekey)).First(&node).Error; err != nil {
 		return node, err
 	}
 	return node, nil
 }
 
+// GetByKey to retrieve full node object from DB or cache, by node_key
+// node_key is expected lowercase
+func (n *NodeManager) GetByKey(nodekey string) (OsqueryNode, error) {
```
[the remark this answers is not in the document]
zhuoyuan-liu: The good thing is that currently, all calling only used the node ID, UUID, and environment ID. So, everything should be good. @javuto  Could you please double-check this part?

zhuoyuan-liu (commented):
@javuto I'd like to get your thoughts on the next steps. If we decide to split the table, it would be a significant breaking change. In that case, we should also add more tests for the package.

pkg/nodes/nodes.go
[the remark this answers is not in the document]
javuto: If we are going to cache the table, then we can explore splitting the table. Otherwise we can just cache a reduced struct with the fields that are used more often (node_key, ID, UUID, etc). Let's decide once we make changes to improve `osctrl-admin` and its UI.

pkg/nodes/utils.go
```diff
@@ -1,49 +1,73 @@
...
+	return n.LastSeen.After(cutoffTime)
 }
 
-// Helper to generate the key to identify a full node in the cache, by node_key
-func CacheFullKeyByNodeKey(n OsqueryNode) string {
-	return CacheFullKeyRaw(n.NodeKey, n.EnvironmentID)
+// ActiveTimeCutoff returns the cutoff time for active nodes
+// based on the specified number of hours
+func ActiveTimeCutoff(hours int64) time.Time {
+	return timeNow().Add(-time.Duration(hours) * time.Hour)
```
zhuoyuan-liu, line 30: Hi @javuto , I forgot to mention that InactiveHours needs to be set to a positive value to make it more intuitive. Otherwise, all nodes will be marked as inactive.

Issue #634, opened 2025-04-03
Race condition for function IncExecution and IncError


zhuoyuan-liu:
Hi @javuto ,

We just see a clear race condition for the following two functions.

The race condition occurs when multiple goroutines call these functions concurrently for the same query.

#### Race Condition Scenario

1. Thread A reads query data: `executions = 5`
2. Thread B reads query data: `executions = 5` (same value, before Thread A updates)
3. Thread A increments and updates: `executions = 6`
4. Thread B increments and updates: `executions = 6` (should be 7, but the increment was lost)

However, this does not impact the core functionality—it only affects the display of the executed query count. Rather than fixing the query function, I recommend completely removing these two functions and using an alternative approach to calculate the number.

```go
// IncExecution to increase the execution count for this query
func (q *Queries) IncExecution(name string, envid uint) error {
	query, err := q.Get(name, envid)
	if err != nil {
		return err
	}
	if err := q.DB.Model(&query).Update("executions", query.Executions+1).Error; err != nil {
		return err
	}
	return nil
}

// IncError to increase the error count for this query
func (q *Queries) IncError(name string, envid uint) error {
	query, err := q.Get(name, envid)
	if err != nil {
		return err
	}
	if err := q.DB.Model(&query).Update("errors", query.Errors+1).Error; err != nil {
		return err
	}
	return nil
}
```

PR #810, opened 2026-05-12
Fix race condition in IncExecution and IncError

kimjune01:
The previous implementation had a race condition where multiple goroutines could read the same counter value, increment it, and write back, causing lost updates under concurrent load.

The read-modify-write cycle allowed this scenario:
1. Thread A reads: executions = 5
2. Thread B reads: executions = 5 (same value)
3. Thread A writes: executions = 6
4. Thread B writes: executions = 6 (should be 7)

Fixed by using atomic database operations with gorm.Expr to perform the increment directly in SQL:

```sql
UPDATE distributed_queries SET executions = executions + 1 WHERE ...
```

This ensures thread-safe atomic increments even under concurrent load, matching the pattern used for IncExpected in the same file.

### DIFF
--- a/pkg/queries/queries.go
+++ b/pkg/queries/queries.go
@@ -439,24 +439,28 @@ func (q *Queries) GetTargets(name string) ([]DistributedQueryTarget, error) {
 
 // IncExecution to increase the execution count for this query
 func (q *Queries) IncExecution(name string, envid uint) error {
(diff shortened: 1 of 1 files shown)


javuto:
Fixes #634 

Thank you!

javuto (approved):

PR #813, opened 2026-05-13
osctrl-api: security hardening — auth bedrock, env secret containment, TLS-side rate-limit

alvarofraguas:
## Summary

Server-side security hardening across the API and TLS surfaces of `osctrl-api` / `osctrl-tls`. Bundled as one PR because the pieces share a coherent threat model (anonymous attack surfaces, cross-tenant exposure, credential-bearing payloads). No API contract changes for existing clients — JSON shapes, route paths, and request bodies are unchanged.

End-to-end tested against a local Kali docker deployment.

### Auth bedrock
- `--auth=jwt` is now the default. `--auth=none` requires the operator to explicitly set `OSCTRL_INSECURE_NO_AUTH=1`; a 60s warning ticker keeps the deployment from drifting into "auth-off forever".
- HttpOnly + Secure cookie session for browser/SPA clients (`osctrl_token`). CLI clients using `Authorization: Bearer` continue to work unchanged.
- Double-submit CSRF (`osctrl_csrf` cookie + `X-CSRF-Token` header) for mutating cookie-authenticated requests. CLI Bearer flows are exempt.
- JWT signing algorithm pinned to HMAC (rejects `alg:none` and RS-vs-HS confusion).
- JWT secret minimum 32 bytes. Startup fails fast with the `openssl rand -base64 48` one-liner if the configured secret is shorter.
- Strict forwarded-headers trust via new `--trusted-proxies` flag. Empty default means `utils.GetIP` ignores `X-Forwarded-For` / `X-Real-IP`, so an internet attacker can't spoof IPs to defeat rate limits or poison audit logs.

### Env secret containment + cross-env defense
- New `types.TLSEnvironmentView` low-privilege projection. Omits `Secret`, `EnrollSecretPath`, `RemoveSecretPath`, `Certificate`, `Flags`, and every other field that materially contributes to enrolling a node.
- `EnvironmentHandler` branches on access level: AdminLevel (or super-admin) gets the full storage struct, UserLevel gets the safe projection.
- `EnvEnrollHandler` / `EnvRemoveHandler` raised from UserLevel to AdminLevel — both embed the env's enroll/remove secret in their responses.
- `EnvActionsHandler` `create`/`delete` branches validate caller-supplied UUIDs via `EnvUUIDFilter` and `ExistsByUUID`.
- `QueryResultsHandler` now precheck-validates the named query belongs to the path env. Without this, a user with QueryLevel on env A could pull results from env B by passing B's query name in A's URL.

### TLS-side hardening
- New `pkg/ratelimit` token-bucket middleware with per-key (IP, IP+env, etc.) buckets, amortized eviction, and a shared overflow bucket past the per-key cap so memory stays bounded under spray.
- Per-(remote, env) rate limits applied to enroll + config endpoints (`--rate-limit-rps`, `--rate-limit-burst`, `--rate-limit-window`).
- Constant-time comparison for env enroll-secret to defeat timing oracles.
- Two new audit hooks: `AuditLog.FailedLogin` (api-side) and `AuditLog.FailedEnroll` (tls-side) so SOC tooling has a clean stream for brute-force / password-spray / enroll-abuse alerts.

### SQL hardening + carve path safety
- New `carves.ValidCarvePath` regexp gate. Without this a `CarveLevel` operator could pass `'; SELECT 1; --` and pivot "carve a file" into "run any SELECT against your fleet" via `GenCarveQuery`'s string-concat path.

### Authz + audit-log hardening
- bcrypt cost raised from default (10) to 12. `CheckLoginCredentials` opportunistically re-hashes existing users at next login — no password reset required, rehash failure is non-fatal.
- New `UserManager.ClearToken` empties `APIToken` AND `CSRFToken` together so an existing JWT+CSRF pair stops validating in lockstep.
- Settings + env-action handlers tightened (env-scoped settings reads instead of global; env service-name validation).
- `pkg/environments/env-cache.go`: amortized cache eviction (dropped a dead field; cache entries TTL out instead of growing unbounded).

### Defaults + ops
- `--audit-log` default flipped to `true` across `deploy/config/{api,admin,tls}.yml`. Operators can opt out with `--audit-log=false`.

## :warning: Operator-visible behavior changes

Two intentional breaking changes worth a release-note line:

1. **`--auth=jwt` is now the default.** Deployments currently running `--auth=none` will fail to start unless they explicitly set `OSCTRL_INSECURE_NO_AUTH=1`. The help text and a 60s warning ticker make this loud. Intended forcing-function so insecure defaults don't survive an upgrade.
2. **bcrypt cost 12 + on-login auto-rehash.** Existing cost-10 hashes migrate transparently the next time each user logs in. No password reset needed, no DB migration. Login latency for affected users will tick up by ~tens of ms once, then stay at the new cost.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./...` — all packages pass
- [x] `gofmt -l ./...` — empty
- [x] End-to-end smoke against a Kali docker deployment (login, CSRF, enroll, env-secret-containment regression)
- [x] Pre-existing tests still pass (no behavior changes on the happy path for existing CLI Bearer clients)
- [x] New tests added:
  - `cmd/api/auth_test.go` — JSON-vs-redirect on 401
  - `cmd/api/handlers/environments_test.go` — env-projection strips every secret-bearing field
  - `pkg/carves/utils_test.go` — `ValidCarvePath` accept/reject table + carve-query happy-path shape
  - `pkg/ratelimit/ratelimit_test.go` — token-bucket + eviction + overflow

## What this is part of

This is round 1 of a 3-round contribution. Round 2 will add the new API endpoints the SPA needs (paginated lists, stats, saved-queries CRUD, users/permissions/tokens, env-config PATCH, audit-log filters). Round 3 will land the React frontend under a new `frontend/` directory. Splitting reduces review surface per PR.


cmd/tls/main.go
```diff
@@ -270,6 +282,16 @@ func osctrlService() {
 			}
 		}()
 	}
+	// Initialize audit log so failed enroll attempts are recorded.
+	//
+	auditLog, err := auditlog.CreateAuditLogManager(db.Conn, serviceName, flagParams.Service.AuditLog)
```
javuto: Let's separate all changes for `osctrl-tls` in its own PR, so it is easier to isolate if something needs to be tested.

pkg/environments/env-cache.go
```diff
@@ -9,6 +9,22 @@ import (
...
+	// the window during which enroll-secret rotations, env deletions, or
+	// config-PATCH changes can be served stale by osctrl-tls — at the
+	// cost of one DB read per cache miss per env per TTL.
+	//
+	// 60s strikes the right balance: security rotations propagate
+	// quickly enough that operators don't need to coordinate a full
+	// osctrl-tls restart, and the steady-state load is "one read per
+	// minute per env" — trivial for osctrl-tls's read traffic.
+	//
+	envCacheTTL = 60 * time.Second
```
javuto: Too low, I prefer to leave it at 2 hours.

pkg/config/flags.go
```diff
@@ -194,8 +194,8 @@ func initServiceFlags(params *ServiceParameters) []cli.Flag {
 		&cli.StringFlag{
 			Name:        "auth",
 			Aliases:     []string{"A"},
-			Value:       AuthNone,
-			Usage:       "Authentication mechanism for the service",
+			Value:       AuthJWT,
+			Usage:       "Authentication mechanism for the service (jwt|none — `none` requires OSCTRL_INSECURE_NO_AUTH=1)",
```
javuto, line 198: Leave usage as `Authentication mechanism for the service`, the rest is invalid.

pkg/ratelimit/ratelimit.go
```diff
@@ -0,0 +1,144 @@
+// Package ratelimit provides a small token-bucket rate-limit middleware
```
javuto, line 1: Same as mentioned in `tls`, these better be in a separate PR for `osctrl-tls`

pkg/carves/utils.go
```diff
@@ -78,7 +79,35 @@ func GenCarveName() string {
...
+// path. The carve string is concatenated into the osquery SQL that
+// every targeted node executes; without this gate a CarveLevel
+// operator could inject arbitrary osquery (e.g. `'; SELECT 1; --`) and
+// pivot from "exfil this path" to "run any SELECT against your nodes".
+//
+// The character class covers realistic carve targets across the three
+// platforms: absolute POSIX paths (Linux/macOS), Windows paths with
+// backslashes and drive letters, and glob wildcards (* and ?). It
+// explicitly excludes single quote, semicolon, and comment markers.
+var validCarvePath = regexp.MustCompile(`^[/A-Za-z0-9._\-\\:*?]+$`)
```
javuto, line 92: Not sure this regex is a good idea. It will block legitimate paths with spaces, for example:
- `C:\Program Files\...`
- `/Library/Application Support/...`

Escaping SQL string literals would be a better approach

javuto (changes requested):
Nice work, several things to change and we can merge!

pkg/config/flags.go
alvarofraguas: Quick confirmation needed before I revert the Usage text: the same diff also changes `Value: AuthNone` → `Value: AuthJWT` on the line above (defaulting the API to JWT auth instead of no-auth, gated by `OSCTRL_INSECURE_NO_AUTH=1` for local dev). Your comment is on the Usage string only — do you want me to keep the default change at `AuthJWT`, or revert the default to `AuthNone` as well?

Context for the default change: it closes the "API is unauthenticated by default" finding from the recent security audit (the related `guardAuthMode` helper in `cmd/api/main.go` is its runtime enforcement). Happy to revert if you'd prefer to land the default change as a separate, narrower PR.

pkg/environments/env-cache.go
alvarofraguas: Reverted to `2 * time.Hour` in c9e73976. Kept the constant + docstring so the value is self-documenting and tunable, but the runtime behavior matches the historical cleanup interval. PRs #814 and #815 rebased onto the new round-1 tip.

cmd/tls/main.go
alvarofraguas: Done — carved the osctrl-tls changes (cmd/tls/handlers/{handlers,post,utils}.go, cmd/tls/main.go, deploy/config/tls.yml) into their own PR: #816. This PR's diff drops to 29 osctrl-api + shared-infra files (no cmd/tls/ touches).

The shared infra they consume (`pkg/auditlog.FailedEnroll`, `pkg/ratelimit`, `utils.SetTrustedProxies` / updated `GetIP`) stays here, since osctrl-api also uses it. #816 stacks on this PR — its diff will collapse to just the 5 TLS files once this lands. PRs #814 and #815 rebased onto the new tip.

pkg/ratelimit/ratelimit.go
alvarofraguas: Quick clarification before I move `pkg/ratelimit`: this PR also wires it into osctrl-api on the `/login` endpoint (10 req/min strict for credential POSTs, separate 60 req/min for pre-auth GETs like the env list). The package itself has two consumers — osctrl-api here and osctrl-tls in #816.

Two options:

**(a)** Keep `pkg/ratelimit` here as shared infrastructure (current state) and only the *consumer* code for osctrl-tls lives in #816. The package is then symmetric with `pkg/auditlog` (also shared, with both `FailedLogin` and `FailedEnroll` helpers).

**(b)** Move `pkg/ratelimit` entirely to #816 and drop the `/login` rate-limit from osctrl-api in this PR. The brute-force / password-spray protection on `/login` would then either ship later in a follow-up or not at all.

Which would you prefer? If (b), happy to land the api-side `/login` rate-limit as a separate small PR after these two settle.

pkg/carves/utils.go
alvarofraguas: Agreed — the regex is too restrictive. Spaces, parentheses, accented characters, and a bunch of legitimate path content fail it. Will replace the validation with SQL string-literal escaping in `GenCarveQuery` (`'` → `''` so a single quote can never break out of the string literal), drop `ValidCarvePath` / `validCarvePath`, and remove the precheck call in `CarvesRunHandler`.

For the glob (LIKE) path I'll handle the wildcard mapping the existing comment already implies — `*` → `%`, `?` → `_` — and escape any user-supplied `%` / `_` / `\\` so a path containing them isn't accidentally interpreted as a wildcard.

Will batch this with your other comments so we don't ping-pong force-pushes on the PR. Tests will cover at least: paths with spaces (Windows and macOS), an injection attempt (`'; SELECT 1; --`), and glob escape of literal `%`.

pkg/ratelimit/ratelimit.go
javuto: We can do (a) and since it is limited to `osctrl-api`, the pkg will be used later by `osctrl-tls` or whatever. Thanks!

alvarofraguas: 👍 Confirmed — keeping `pkg/ratelimit` here as shared infra. No code change needed; the osctrl-tls consumer is already split out into #816 as you requested.

pkg/config/flags.go
alvarofraguas: Got it — you're right, and I missed that. The `--auth` flag is in the shared `initServiceFlags` so my default flip applied to all three services, including `osctrl-tls`. osctrl-tls's runtime never reads the field (osquery agents authenticate via the per-environment enroll secret, not JWT), so the practical impact is zero, but `osctrl-tls --help` now misleadingly shows `auth (default: jwt)` which is wrong semantically.

Proposing the cleanest fix: split the `--auth` flag out of the shared `initServiceFlags` into the three per-service `Init*Flags` functions, with per-service defaults:

- `osctrl-api` → `AuthJWT` (closes U-AUTH-1 — silent unauthenticated default was the audit blocker)
- `osctrl-admin` → `AuthDB` (its existing default, browser login form)
- `osctrl-tls` → `AuthNone` (correct for it, as you said)

Plus the Usage text reverted to `Authentication mechanism for the service` per your other comment.

The `guardAuthMode` helper stays in `cmd/api/main.go` only — it's specific to osctrl-api's audit finding. Want me to land this as part of #813, or as a small standalone PR you can review independently?

javuto: Okay understood. Then let's leave it like this, including the usage.

javuto (approved):

PR #814, opened 2026-05-13
osctrl-api: API extensions for a React admin frontend (round 2 of 3)

alvarofraguas:
## Summary

Round 2 of 3. Adds the API surface a React admin frontend needs to fully cover what the current `osctrl-admin` templates do. Every new endpoint is **additive** — no existing routes are removed or repurposed. The legacy admin keeps working unchanged.

End-to-end tested against a Kali docker deployment.

:warning: **Stacked on #813** (security hardening). Merge #813 first, then this. If #813 is merged before review starts here, this branch will be cleanly rebased on the new `main` HEAD with no conflicts.

### Stats / dashboard
| Method | Path | Purpose |
|--------|------|---------|
| GET | `/api/v1/stats` | Cross-env summary KPIs |
| GET | `/api/v1/stats/osquery-versions` | Fleet agent versions |
| GET | `/api/v1/stats/activity/{env}` | Env-scoped audit-log activity heatmap |
| GET | `/api/v1/stats/activity/node/{env}/{uuid}` | Per-node activity heatmap |
| GET | `/api/v1/stats/activity/node-batch/{env}` | Per-node heatmap, up to 100 uuids |

### Logs (live log viewer)
| Method | Path | Purpose |
|--------|------|---------|
| GET | `/api/v1/logs/{type}/{env}/{uuid}` | Paginated, since-aware, optional `?q=` search |

### Saved queries (full CRUD)
| Method | Path |
|--------|------|
| GET | `/api/v1/saved-queries/{env}` |
| POST | `/api/v1/saved-queries/{env}` |
| PATCH | `/api/v1/saved-queries/{env}/{name}` |
| DELETE | `/api/v1/saved-queries/{env}/{name}` |

### User profile + permissions + token
| Method | Path |
|--------|------|
| GET | `/api/v1/users/me` |
| PATCH | `/api/v1/users/me` |
| POST | `/api/v1/users/me/password` |
| POST | `/api/v1/users/{username}/permissions` |
| POST | `/api/v1/users/{username}/token/refresh` |
| DELETE | `/api/v1/users/{username}/token` |

### Environment CRUD + config PATCHes
| Method | Path |
|--------|------|
| POST | `/api/v1/environments` |
| PATCH | `/api/v1/environments/{env}` |
| DELETE | `/api/v1/environments/{env}` |
| GET | `/api/v1/environments/{env}/config` |
| PATCH | `/api/v1/environments/{env}/config` |
| PATCH | `/api/v1/environments/{env}/intervals` |
| PATCH | `/api/v1/environments/{env}/expiration` |

### Settings PATCH
| Method | Path |
|--------|------|
| PATCH | `/api/v1/settings/{service}/{name}` |

### Audit log filters + pagination
| Method | Path |
|--------|------|
| GET | `/api/v1/audit-logs?service=&username=&type=&envUuid=&since=&until=&page=&pageSize=` |

### Login envs (pre-auth env list for the SPA's login page env picker)
| Method | Path |
|--------|------|
| GET | `/api/v1/login/environments` (UUID + name only; everything else stays behind auth) |

### Sample libraries (operator starter packs)
| Method | Path |
|--------|------|
| GET | `/api/v1/queries/samples` |
| GET | `/api/v1/carves/samples` |
| GET | `/api/v1/osquery/tables` |

## Pagination, sort, search conventions

Every list endpoint accepts `?page=&page_size=` (default 50, max 500) and returns the envelope:

```json
{ "items": [...], "page": N, "page_size": N, "total_items": N, "total_pages": N }
```

Sortable fields use a per-resource `SortableColumns` allowlist enforced at the package layer (`pkg/nodes`, `pkg/queries`, `pkg/carves`). Unknown sort keys fall back to the resource's default order without 400ing.

Search is `?q=` free-text against a per-resource field set (case-insensitive LIKE). Wildcards are escaped server-side.

## New package: `pkg/dbutil`

Dialect-aware SQL bucket-expression helper (postgres / mysql / sqlite) used by the activity heatmap endpoints. Each category (status logs / result logs / distributed queries / carves) issues a single SQL `GROUP BY` rather than plucking every timestamp — at 50k+ nodes the table-page heatmap query is bounded by the index instead of the chatty-row count.

## Package-layer additions

- **`pkg/nodes`** — `GetByEnvPaged`, `NodeView` projection, `SortableColumns`, platform-bucket helpers, `GetOsqueryVersionCounts`.
- **`pkg/queries`** — `GetByEnvTargetPaged`, `GetSaved*` CRUD, `SortableColumns`, sample-template loader, `GetNodeQueryBucketed`.
- **`pkg/carves`** — `GetByEnvPaged`, sample-template loader, `GetNodeCarveBucketed`.
- **`pkg/environments`** — `Create` / `Update` / `Delete`, `UpdateConfig` / `UpdateIntervals` / `UpdateExpiration` helpers.
- **`pkg/auditlog`** — `GetPaged` with `PageFilter`; `GetEnvActivityBucketed` for the heatmap.
- **`pkg/logging`** — `GetNodeLogs` with `?q=` search filter, `GetNode{Status,Result}Bucketed` for the heatmap.
- **`pkg/osquery`** — `LoadTables` (osquery schema for the SPA query editor).
- **`pkg/types`** — `NodeView`, paginated response envelopes, `EnvCreate` / `EnvUpdate` / `EnvConfig*` request types, `SettingPatchRequest`, `SavedQueryView`, `AdminUserView`.

## Test plan

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `go test ./...` — all packages pass
- [x] `gofmt -l ./...` — empty
- [x] End-to-end smoke against a Kali docker deployment (paginated lists, activity heatmap, env CRUD, saved-queries CRUD, profile flow)
- [x] New tests added: `cmd/api/handlers/stats_test.go`, `pkg/nodes/nodes_test.go`, `pkg/queries/queries_test.go`, `pkg/queries/saved_test.go`

## What this enables

Round 3 will land the React admin SPA under a new `frontend/` directory at the repo root. The SPA consumes only the endpoints in this PR — no admin-template surface is touched.

### DIFF
--- a/cmd/api/handlers/environments_crud.go
+++ b/cmd/api/handlers/environments_crud.go
@@ -35,7 +35,11 @@ func (h *HandlersApi) EnvironmentCreateHandler(w http.ResponseWriter, r *http.Re
 		apiErrorResponse(w, "error parsing POST body", http.StatusBadRequest, err)
 		return
 	}
-	body.Name = strings.TrimSpace(body.Name)
+	// Names are persisted lowercase so `Dev` and `dev` collide on the
+	// uniqueIndex and so name-based lookups across the codebase don't
+	// need to special-case casing. Filter validation runs against the
+	// canonical form.
+	body.Name = strings.ToLower(strings.TrimSpace(body.Name))
 	body.Hostname = strings.TrimSpace(body.Hostname)
 	if !environments.VerifyEnvFilters(body.Name, body.Icon, body.Type, body.Hostname) {
 		apiErrorResponse(w, "invalid name, hostname, type, or icon", http.StatusBadRequest, nil)
@@ -60,6 +64,15 @@ func (h *HandlersApi) EnvironmentCreateHandler(w http.ResponseWriter, r *http.Re
 	}
 	env.Flags = flags
 	if err := h.Envs.Create(&env); err != nil {
+		// Race fallback: the Exists() check above is best-effort; if a
+		// concurrent request inserted the same name between that check
+		// and this Create, the uniqueIndex on tls_environments.name
+		// raises gorm.ErrDuplicatedKey. Return 409 instead of 500 so
+		// the SPA can distinguish the collision from a real DB error.
+		if errors.Is(err, gorm.ErrDuplicatedKey) {
+			apiErrorResponse(w, "environment with that name already exists", http.StatusConflict, err)
+			return
+		}
 		apiErrorResponse(w, "error creating environment", http.StatusInternalServerError, err)
 		return
 	}
@@ -133,16 +146,19 @@ func (h *HandlersApi) EnvironmentUpdateHandler(w http.ResponseWriter, r *http.Re
 	//
 	patch := map[string]interface{}{}
 	if body.Name != nil {
-		n := strings.TrimSpace(*body.Name)
+		// Persist env names lowercased — see EnvironmentCreateHandler for
+		// the reasoning. Filter validation runs against the canonical
+		// form so the regex character-class still matches.
+		n := strings.ToLower(strings.TrimSpace(*body.Name))
 		if !environments.EnvNameFilter(n) {
 			apiErrorResponse(w, "invalid environment name", http.StatusBadRequest, fmt.Errorf("rejected name %q", *body.Name))
 			return
 		}
 		if n != env.Name {
-			// Reject a rename that would collide with an existing env.
-			// Mirrors the create-path uniqueness check; without this gate
-			// PATCH would silently produce two environments with the same
-			// name, which downstream lookups by name handle inconsistently.
+			// Pre-flight: reject a rename that would collide with an
+			// existing env (best-effort, common case). The uniqueIndex
+			// on tls_environments.name is the actual source of truth and
+			// catches the TOCTOU race below.
 			if h.Envs.Exists(n) {
 				apiErrorResponse(w, "environment already exists", http.StatusConflict, fmt.Errorf("environment %s already exists", n))
 				return
@@ -188,6 +204,15 @@ func (h *HandlersApi) EnvironmentUpdateHandler(w http.ResponseWriter, r *http.Re
 		return
 	}
 	if err := h.Envs.DB.Model(&env).Updates(patch).Error; err != nil {
+		// Race fallback for the rename path: a concurrent PATCH could
+		// have inserted the same target name between our Exists() check
+		// above and this UPDATE. uniqueIndex catches it; translate to
+		// 409 so the SPA can distinguish the collision from a real DB
+		// error.
+		if errors.Is(err, gorm.ErrDuplicatedKey) {
+			apiErrorResponse(w, "environment already exists", http.StatusConflict, err)
+			return
+		}
 		apiErrorResponse(w, "error updating environment", http.StatusInternalServerError, err)
 		return
 	}
@@ -304,7 +329,14 @@ func (h *HandlersApi) EnvironmentConfigPatchHandler(w http.ResponseWriter, r *ht
 		apiErrorResponse(w, "error parsing PATCH body", http.StatusBadRequest, err)
 		return
 	}
-	// Validate every supplied section is parseable JSON before writing any.
+	// Validate every supplied section is parseable JSON, and canonicalize
+	// each value (trim whitespace, "" → "{}") into `normalized` so the
+	// persistence calls below share the same source of truth as the
+	// validation. Without this, the validator accepts an empty string
+	// (treating it as "{}") while the update calls below would still
+	// write the original empty string to the DB — downstream parsers
+	// (osquery config generation, admin JSON view) trip on the empty
+	// string and behave inconsistently with the validated shape.
 	sections := map[string]*string{
 		"options":    body.Options,
 		"schedule":   body.Schedule,
@@ -313,11 +345,11 @@ func (h *HandlersApi) EnvironmentConfigPatchHandler(w http.ResponseWriter, r *ht
 		"atc":        body.ATC,
 		"flags":      body.Flags,
 	}
+	normalized := map[string]string{}
 	for name, val := range sections {
 		if val == nil {
 			continue
 		}
-		// Empty string isn't valid JSON; treat as the empty object.
 		s := strings.TrimSpace(*val)
 		if s == "" {
 			s = "{}"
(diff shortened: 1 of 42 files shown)


cmd/api/handlers/environments_crud.go
```diff
@@ -0,0 +1,506 @@
...
+	// route paths) would happily embed unescaped.
+	//
+	patch := map[string]interface{}{}
+	if body.Name != nil {
+		n := strings.TrimSpace(*body.Name)
+		if !environments.EnvNameFilter(n) {
+			apiErrorResponse(w, "invalid environment name", http.StatusBadRequest, fmt.Errorf("rejected name %q", *body.Name))
+			return
+		}
+		if n != env.Name {
```
javuto, line 157: There is no actual check whether if the environment already exists (by name). On environment creation this check is performed.

alvarofraguas: Good catch — fixed in d18878c1.

PATCH now calls `h.Envs.Exists(n)` before accepting a rename, mirroring the create-path check. Returns **409 Conflict** rather than 400, since this is a uniqueness conflict (the create handler still returns 400 for consistency with its older shape, but I felt 409 was the more accurate REST semantic for the PATCH case — happy to align both on 400 if you'd prefer).

cmd/api/handlers/environments_crud.go
```diff
@@ -0,0 +1,506 @@
...
+		if s == "" {
+			s = "{}"
+		}
+		var probe interface{}
+		if err := json.Unmarshal([]byte(s), &probe); err != nil {
+			apiErrorResponse(w, fmt.Sprintf("section %q is not valid JSON: %s", name, err.Error()), http.StatusBadRequest, err)
+			return
+		}
+	}
+	if body.Options != nil {
```
javuto: For each of these sections, if they have an empty strings, the value `{}` should be persisted, like done above for validation, otherwise invalid values are actually saved.

javuto (changes requested):

cmd/api/handlers/environments_crud.go
alvarofraguas: Quick context before I push more: the `Exists()` check I added in d18878c1 closes the single-request case but doesn't survive a real race — two concurrent PATCHes can both pass the existence check and then both succeed at the UPDATE, ending with two rows that share a name. Looked into properly fixing it and found a few related items worth surfacing.

What I want to add (working on it now):

1. **DB-level UNIQUE constraint on `tls_environments.name`** via AutoMigrate. Defends against the race regardless of which handler does the write (PATCH, create, future bulk import, the CLI).
2. **Enable `gorm.Config{TranslateError: true}`** in `pkg/backend` so handlers can use `errors.Is(err, gorm.ErrDuplicatedKey)` cleanly across the three supported dialects.
3. **Update `pkg/queries/saved.go`** which does a string-match on `"duplicate key"` / `"Duplicate entry"` / `"UNIQUE constraint"` today — needs to switch to `errors.Is` once TranslateError is on, since the driver translator replaces the raw message.
4. **Both create + PATCH handlers**: catch `ErrDuplicatedKey` from the actual `INSERT`/`UPDATE` as a fallback to the pre-flight `Exists` check, return 409 in both paths.
5. **Concurrency test** in `pkg/environments` that races two writes and asserts exactly one wins.

Two questions for you before I implement:

**Q1 — case sensitivity.** `EnvNameFilter` allows `[a-zA-Z0-9_-]+`, so `Dev` and `dev` are both valid. With a plain unique index they're distinct; with `LOWER(name)` they collide. My read: operators probably think of them as the same env, so case-insensitive feels right. Your call?

**Q2 — existing data.** Are you aware of any production deployment that has duplicate env names today? AutoMigrate adding a unique constraint against a table with duplicates fails fatally, which breaks boot until cleaned up. If you've never seen it in the wild, fine; if you have, the migration needs a pre-check + clear error.

I'll hold the rest of #814 (including the `Exists()` fix from d18878c1) until you weigh in, then ship the full thing in one coherent commit so the history is clean.

cmd/api/handlers/environments_crud.go
alvarofraguas: Confirmed bug — good catch. The empty-string-to-`{}` normalization at line 322–324 lives only inside the validation probe; the actual `UpdateOptions` / `UpdateSchedule` / `UpdatePacks` / `UpdateDecorators` / `UpdateATC` / `UpdateFlags` calls below use `*body.Options` / `*body.Schedule` / etc., which still hold the original empty string. Empty strings round-trip into the DB as `""` rather than `{}`, and downstream parsers (osquery config generation, the admin UI's JSON view) trip on them.

Will fix by hoisting the normalized values into a parallel map and having the update calls read from that, so validation + persistence share one source of truth.

Batching this with the env-uniqueness fix from the other thread — will push one coherent commit for #814 once we settle Q1 (case sensitivity) and Q2 (existing-data assumption). Full verification (real concurrency race test + Kali end-to-end smoke for both fixes) before push.

cmd/api/handlers/environments_crud.go
javuto: Yeah, `Dev` and `dev` should be the same. Environment name should be lowercased to avoid that edge case. Regarding Q2, I don't think we should worry about it.

alvarofraguas: Done in 4ac71b84. Both threads addressed in one commit.

**Env-name uniqueness** (full TOCTOU-safe fix):
- UNIQUE index on `tls_environments.name` via AutoMigrate.
- `TranslateError: true` in pkg/backend so handlers use `errors.Is(err, gorm.ErrDuplicatedKey)`.
- `nameRegex` tightened to lowercase-only `[a-z0-9_-]+`; create + PATCH handlers lowercase input before validation (friendly UX, canonical storage).
- Race fallback in both handlers: the actual `Create`/`Updates` catches `ErrDuplicatedKey` → 409 instead of 500.

**JSON config normalization** (line :323):
- Empty-string sections now get persisted as `{}` (was: validated as `{}` but stored as `""`). Fixed by hoisting normalized values into a parallel map shared by validation + persistence.

**Tests added** (`pkg/environments/uniqueness_test.go`):
- Single-insert dup → `ErrDuplicatedKey`
- Case-sensitivity at storage level (locked as contract)
- UPDATE-rename collision → `ErrDuplicatedKey`
- 16-goroutine concurrent-insert race → exactly 1 winner + 15 dup-key losers + 0 unclassified errors

**Verified end-to-end on a real Postgres dev stack**: 14 curl tests covering create / PATCH / config-patch / two parallel curls racing the same rename. The race scenario (the gap d18878c1 left open) returns one HTTP 200 and one HTTP 409 with the right error message; DB ends with one row. Full transcript on request.

cmd/api/handlers/environments_crud.go
alvarofraguas: Also addressed in 4ac71b84 — same commit since it touches the same file. Hoisted the canonicalized values into a `normalized map[string]string` so the validation probe and the update calls below share one source of truth. Verified end-to-end against Postgres: PATCH config with `{"options":"","schedule":"   ","packs":"{\"my_pack\":1}","decorators":""}` ends up with `options='{}', schedule='{}', packs='{"my_pack":1}', decorators='{}'` in the DB. Invalid JSON still rejected with 400 + section name + parse error.

cmd/api/handlers/environments_crud.go
alvarofraguas: One heads-up on legacy-admin interactions. The legacy admin code in `cmd/admin/` wasn't touched in this PR (it's out of scope), but the schema and filter changes here do affect what the legacy admin sees at runtime. Surfacing so you have the full picture, not asking for anything:

- Legacy admin "create env" form: rejecting uppercase names via the tightened `EnvNameFilter` regex. Operator submitting `Prod` gets the existing "invalid data" 400 response; they have to retry with `prod`. The legacy admin process keeps running fine; this is the same error shape as any other validation rejection.
- Legacy admin "create env" concurrent race: if two operators submit the same name at the same time and the `Exists()` check races, the loser now gets a 500 "error creating environment" instead of silently producing a duplicate row. Replaces a silent data-corruption bug with a loud error.
- Legacy admin "delete env" form: any pre-existing env with an uppercase name in the DB can no longer be deleted via this form (the delete path also calls `EnvNameFilter`). `osctrl-cli env remove --name <whatever>` still works as the cleanup path.

If you want me to touch `cmd/admin/handlers/post.go` to either (a) lowercase the create-path input, or (b) loosen the regex used by the delete path so legacy uppercase data can still be removed through the form, happy to do it in a small follow-up — just out of scope of this PR as I read it.

javuto (approved):

PR #824, opened 2026-05-18
docker-compose standalone is not the standard way to trigger docker c…

ITJamie:
…ompose anymore

`docker compose` is the modern way of triggering docker compose files


javuto (approved):
Thanks for the PR!

PR #827, opened 2026-05-18
Fix for goreleaser error when tagging a new release

javuto:
Potential fix for tagged release errors using GoReleaser: Example https://github.com/jmpsec/osctrl/actions/runs/26044376278


alvarofraguas:
Discussed offline — also need a `dockers:` + `docker_manifests:` split to address the manifest-list export error that's been failing the workflow on every tagged release. Coming in a follow-up.

PR #828, opened 2026-05-18
release: fix multi-arch Docker builds (manifest list export error)

alvarofraguas:
## Summary

- Every tagged release since v0.4.8 fails at the Docker step with:
  ```
  docker exporter does not currently support exporting manifest lists
  ```
  The root cause is that a single `buildx` invocation with `--platform=linux/amd64,linux/arm64` tries to export a manifest list, which the default Docker exporter does not support.

- **Fix:** split each component's `dockers:` block into two per-arch entries (e.g. `osctrl-tls:<version>-amd64` and `osctrl-tls:<version>-arm64`), then combine them via `docker_manifests:` into the final multi-arch tags (`<version>` and `latest`). `docker pull <image>:<tag>` now resolves the correct arch automatically.

- `skip_push` on manifest entries is gated by `{{ .IsSnapshot }}` so snapshot builds still work without pushing.

## Changes

- `.goreleaser.yml`: 4 multi-platform docker blocks → 8 per-arch blocks + 8 `docker_manifests` entries

## Test plan

- [x] `goreleaser check` validates the config
- [ ] Tag a pre-release and verify Docker images are pushed with correct multi-arch manifests

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -121,56 +121,159 @@ changelog:
       - Merge branch
 
 dockers:
+  # osctrl-tls — amd64
   - image_templates:
-      - "{{ .Env.DOCKER_HUB_ORG }}/osctrl-tls:{{ .Version }}"
-      - "{{ .Env.DOCKER_HUB_ORG }}/osctrl-tls:latest"
+      - "{{ .Env.DOCKER_HUB_ORG }}/osctrl-tls:{{ .Version }}-amd64"
     dockerfile: deploy/cicd/docker/Dockerfile-osctrl-tls
     use: buildx
+    goarch: amd64
     skip_push: false
     build_flag_templates:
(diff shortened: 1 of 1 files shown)


javuto (approved):
👏

alvarofraguas:
The main CI run triggered by this merge ([26082810103](https://github.com/jmpsec/osctrl/actions/runs/26082810103/job/76688472466)) got past the manifest-list error but hit a second bug: the arm64 docker builds fail with `"/osctrl-admin-linux-amd64": not found`.

**Root cause:** The Dockerfiles declare `ARG GOARCH=amd64` as default, and the arm64 goreleaser blocks pass `--platform=linux/arm64` but never override `GOARCH`. So the `COPY` instruction looks for the amd64 binary in an arm64 build context.

Fix in #839 — adds `--build-arg=GOARCH=arm64` to all four arm64 docker blocks.

PR #837, opened 2026-05-19
api: fine-grained permission control for service account creation

### LINKED ISSUE #774: Add fine-grained permission control for service accounts
## Problem

Currently, when creating service accounts via the API, permissions are tied to the global admin flag. This means users can only create accounts with either:
- Full admin access (admin=true) - all permissions to all environments
- Read-only access (admin=false) - only read permission

There is no way to create service accounts with specific combinations of permissions like:
- Read + Query (for monitoring tools)
- Read + Carve (for forensics tools)
- Read + Query + Carve + Admin (for environment-specific admins)

## Solution

Implement fine-grained permission control by adding individual permission flags to the API user creation endpoint:
- `user_access` - Read access (view nodes, queries, environments, etc.)
- `query_access` - Query execution access
- `carve_access` - File carving access
- `admin_access` - Admin access for specified environments

## Benefits

1. **Security**: Principle of least privilege - give accounts only the permissions they need
2. **Flexibility**: Support various use cases (monitoring, forensics, CI/CD, etc.)
3. **Backward Compatibility**: Existing API clients continue to work without changes
4. **Better Service Account Management**: Create purpose-specific accounts

## Implementation

- [ ] Add permission fields to `ApiUserRequest` type
- [ ] Update `UserActionHandler` to use fine-grained permissions
- [ ] Update Swagger documentation
- [ ] Add comprehensive testing tools
- [ ] Maintain backward compatibility

## Testing Tools

Added testing scripts in `tools/testing/`:
- Create service accounts with random permissions
- Validate permissions via API calls
- Export credentials to CSV
- Debug API connectivity

## API Usage

```bash
POST /api/v1/users/my-service-account/add
{
  "username": "my-service-account",
  "password": "secure_password",
  "email": "service@example.com",
  "service": true,
  "admin": false,
  "environments": ["env-uuid"],
  "user_access": true,
  "query_access": true,
  "carve_access": true,
  "admin_access": false
}
```

alvarofraguas:
## Summary

- Adds optional `user_access`, `query_access`, `carve_access`, and `admin_access` fields to the user creation API endpoint
- When provided, these override the legacy behavior where all permission levels were derived from the binary `admin` flag
- Enables creating service accounts with specific permission combinations (e.g. read + query for monitoring, read + carve for forensics)

## Backward compatibility

Fields use pointer types (`*bool`) so `nil` (not provided) is distinguishable from `false` (explicitly denied). When none of the new fields are provided, the handler falls back to the existing behavior: `user_access=true`, everything else = `admin` value.

Existing API clients and the CLI continue working unchanged.

## API usage

```bash
POST /api/v1/users/my-monitoring-bot/add
{
  "username": "my-monitoring-bot",
  "password": "secure_password",
  "email": "monitoring@example.com",
  "service": true,
  "admin": false,
  "environments": ["env-uuid"],
  "user_access": true,
  "query_access": true,
  "carve_access": false,
  "admin_access": false
}
```

## Security

The endpoint is already gated behind global admin permission (`AdminLevel` check). Only admins can create users or assign permissions. This change adds granularity to what admins can already do — it does not expand the attack surface.

Closes #774

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./cmd/api/... ./pkg/users/...` passes
- [x] Legacy behavior preserved: omitting new fields produces same permissions as before
- [x] Fine-grained: `admin=false` + `query_access=true` creates user with read + query only

### DIFF
--- a/cmd/api/handlers/users.go
+++ b/cmd/api/handlers/users.go
@@ -160,7 +160,26 @@ func (h *HandlersApi) UserActionHandler(w http.ResponseWriter, r *http.Request) 
 				return
 			}
 		}
-		access := h.Users.GenEnvUserAccess(envs, true, (u.Admin), (u.Admin), (u.Admin))
+		// Determine per-level access. When fine-grained fields are
+		// provided, use them; otherwise fall back to the legacy
+		// behavior where admin implies all access levels.
+		userAcc := true
+		queryAcc := u.Admin
+		carveAcc := u.Admin
+		adminAcc := u.Admin
+		if u.UserAccess != nil {
+			userAcc = *u.UserAccess
+		}
+		if u.QueryAccess != nil {
+			queryAcc = *u.QueryAccess
+		}
+		if u.CarveAccess != nil {
+			carveAcc = *u.CarveAccess
+		}
+		if u.AdminAccess != nil {
+			adminAcc = *u.AdminAccess
+		}
+		access := h.Users.GenEnvUserAccess(envs, userAcc, queryAcc, carveAcc, adminAcc)
 		perms := h.Users.GenPermissions(u.Username, ctx[ctxUser], access)
 		if err := h.Users.CreatePermissions(perms); err != nil {
 			apiErrorResponse(w, "error creating permissions", http.StatusInternalServerError, err)
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -170,6 +170,10 @@ type ApiUserRequest struct {
 	NotService   bool     `json:"not_service"`
 	API          bool     `json:"api"`
 	Environments []string `json:"environments"`
+	UserAccess   *bool    `json:"user_access,omitempty"`
+	QueryAccess  *bool    `json:"query_access,omitempty"`
+	CarveAccess  *bool    `json:"carve_access,omitempty"`
+	AdminAccess  *bool    `json:"admin_access,omitempty"`
 }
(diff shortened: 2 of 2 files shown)


javuto:
I see a problem with this, when having SAML/OIDC enabled, the email is the username, so it may be better to not have the username passed as the URL, but just be `POST /api/v1/users/add`. What do you think?

alvarofraguas:
Good point — this PR doesn't change the routing though, the `POST /api/v1/users/{username}/{action}` pattern is pre-existing.

Regarding the email-as-username concern: our OIDC/SAML PRs (#829–#833) enforce a strict character-class regex on usernames at the auth boundary:

```go
var usernameAllowed = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,64}$`)
```

This rejects `@` and `.` entirely, so a full email address can never become a username through the federated login flow. The default `username_claim` is `preferred_username` (not `email`), and if someone configures `username_claim: email`, the sanitization rejects it and the login fails. The email is stored separately in `AdminUser.Email`.

That said, the current upstream `cmd/admin/oidc.go` doesn't have this sanitization, so the concern is valid for existing deployments. Happy to also refactor the routing to use `POST /api/v1/users/add` (username only in the body) as a separate cleanup PR if you'd like — it's a cleaner API shape regardless.

javuto (approved):

PR #856, opened 2026-06-12
Reload frontend build when changes to code

javuto:
Same behavior as the other services (`osctrl-tls`, `osctrl-api`, `osctrl-admin`) and reload the service in the Docker development environment when codes detected under `frontend/`.


alvarofraguas (approved):
LGTM

PR #1015, opened 2026-09-02
Reject unknown user actions instead of silently returning 200

mehdi-shokohi:
UserActionHandler switches on the {action} path segment but had no default case. An unrecognized action (e.g. "update") matched none of add/edit/remove, skipped every branch, and still fell through to the trailing HTTPResponse — returning 200 {"data":""} while creating, editing, or removing nothing. Callers reasonably read the 200 as success, so a typo'd action silently no-ops.

Add a default case that returns 400 "invalid action".

### DIFF
--- a/cmd/api/handlers/users.go
+++ b/cmd/api/handlers/users.go
@@ -318,6 +318,13 @@ func (h *HandlersApi) UserActionHandler(w http.ResponseWriter, r *http.Request) 
 			}
 		}
 		returnData = "user removed successfully"
+	default:
+		// Reject unknown actions explicitly. Without this case an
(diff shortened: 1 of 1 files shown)


javuto:
Thanks for the PR! I can amend the signature, but in the future, please do sign your commits. Thanks again!

javuto:
Actually I can not amend it, since it comes from your fork. Can you sign the commit please? Thanks!

javuto (approved):

javuto (changes requested):
Please sign the commit, thanks!

javuto (changes requested):
Please sign the commit to be able to merge, thanks!

javuto (approved):

